PR #31, opened 2025-02-07
[3] E2E stack. Added Nethermind node

Filter94:


gauravahuja (approved):

docker/compose.yaml
```diff
@@ -177,7 +177,7 @@ services:
     depends_on:
       sequencer:
         condition: service_healthy
-    image: erigontech/erigon:v2.61.2
+    image: ethpandaops/erigon:main-7b671790
```
jframe, line 180: why aren't you using the official docker image?

Filter94: Good question. `erigontech/erigon` tagged with `v*` can't start up with the script I have because of:
```
/scripts/node.sh: 10: source: not found
```
`erigontech/erigon` tagged with `main-*` are based on other distributive and they work, but they aren't meant to be long lived and they are deleted in less than a month after release.

`ethpandaops/erigon` `main-*` have the available commands and long lived.

I could adapt the scripts to be compatible with `erigontech/erigon`, but given that Erigon maintenance is not the main focus of this repo, I've decided I'd just use a different image with fairly good support.

jframe (approved):

PR #45, opened 2025-02-24
Switching from scratch

Filter94:


executionlayer/src/main/kotlin/maru/executionlayer/client/ExecutionLayerClient.kt
```diff
@@ -29,12 +28,12 @@ import tech.pegasys.teku.infrastructure.bytes.Bytes8
...
+  fun getPayload(payloadId: Bytes8): SafeFuture<Response<ExecutionPayloadV1>>
 
-  fun newPayload(executionPayload: ExecutionPayloadV3): SafeFuture<Response<PayloadStatusV1>>
+  fun newPayload(executionPayload: ExecutionPayloadV1): SafeFuture<Response<PayloadStatusV1>>
 
   fun forkChoiceUpdate(
     forkChoiceState: ForkChoiceStateV1,
-    payloadAttributes: PayloadAttributesV3?,
+    payloadAttributes: PayloadAttributesV1?,
   ): SafeFuture<Response<ForkChoiceUpdatedResult>>
```
gauravahuja, line 38: Do you think it would be better to add a V1 suffix to the interface name or the method names?
`ExecutionLayerClientV1` or `getPayloadV1` ?

e2e/src/acceptanceTest/kotlin/maru/e2e/TestEnvironment.kt
```diff
@@ -43,7 +43,7 @@ object TestEnvironment {
     )
   val sequencerL2Client: Web3j = buildWeb3Client("http://localhost:8545")
 
-  // The switch doesn't work for Geth 1.14 yet
+  // The switch doesn't work for Geth 1.15 yet
   val geth1L2Client: Web3j = buildWeb3Client("http://localhost:8555")
```
gauravahuja, line 47: It might be better to create constants for all these Port numbers and also add a comment that these port numbers should match the ones in the docker compose.

gauravahuja (commented):
lgtm just the one comment.

executionlayer/src/main/kotlin/maru/executionlayer/client/ExecutionLayerClient.kt
Filter94: I'm not sure yet. I think this interface has to be generic, but I'll think about it in my next PR

e2e/src/acceptanceTest/kotlin/maru/e2e/TestEnvironment.kt
Filter94: This class is meant to be the only piece that is coupled to the docker-compose and these ports aren't meant to be used anywhere else in Kotlin code. What would be the benefit of turning them into constants?

e2e/src/acceptanceTest/kotlin/maru/e2e/CliqueToPosTest.kt
```diff
@@ -63,145 +66,173 @@ class CliqueToPosTest {
...
+        "http://localhost:11551",
+        TestEnvironment.jwtConfig,
+      )
+    private val geth1ExecutionEngineClient = createExecutionClient("http://localhost:8561", TestEnvironment.jwtConfig)
+    private val geth2ExecutionEngineClient = createExecutionClient("http://localhost:8571", TestEnvironment.jwtConfig)
+    private val gethSnapServerExecutionEngineClient =
+      createExecutionClient("http://localhost:8581", TestEnvironment.jwtConfig)
+    private val gethExecutionEngineClients =
+      mapOf(
+//        "follower-geth" to geth1ExecutionEngineClient,
```
jframe, line 100: why did you remove the follower-geth node?

Filter94: Because Geth 1.14+ won't sync with Clique, so it won't switch. I think it should sync from scratch though, so let me try and add it to the full sync test

e2e/src/acceptanceTest/kotlin/maru/e2e/TestEnvironment.kt
Filter94: I added engine API clients here as well, so it's all in the same pile

gauravahuja: These constants are also used in CliqueToPosTest.kt
```
e2e/src/acceptanceTest/kotlin/maru/e2e/CliqueToPosTest.kt
77:  private val web3jClient = createWeb3jClient("http://localhost:8550", Optional.empty())
79:  private val besuFollowerExecutionEngineClient = createExecutionClient("http://localhost:9550")
82:      "http://localhost:10550",
87:      "http://localhost:11551",
90:  private val geth1ExecutionEngineClient = createExecutionClient("http://localhost:8561", TestEnvironment.jwtConfig)
91:  private val geth2ExecutionEngineClient = createExecutionClient("http://localhost:8571", TestEnvironment.jwtConfig)
93:    createExecutionClient("http://localhost:8581", T
```

Filter94: No, these ports are different. These are for the engine API

app/src/test/kotlin/maru/app/MaruDummyConsensusTest.kt
```diff
@@ -133,7 +133,9 @@ class MaruDummyConsensusTest {
     val timestamps = blocks.map { it.timestamp.toLong() }
     (1.until(blocks.size)).forEach {
       assertThat(timestamps[it - 1]).isLessThan(timestamps[it])
-      assertThat(timestamps[it] - timestamps[it - 1]).isEqualTo(blockTimeSeconds)
+      val actualBlockTime = timestamps[it] - timestamps[it - 1]
```
gauravahuja, line 136: It is not clear if toLong on line 133 converts it to seconds or milliseconds but the assert compares it with seconds.

gauravahuja (approved):

app/src/test/kotlin/maru/app/MaruDummyConsensusTest.kt
Filter94: Block timestamps are always in seconds on chain, but I'll change the name for clarity

PR #46, opened 2025-02-25
[43] Add p2p config for boot node and validator nodes

gauravahuja:


config/src/main/kotlin/maru.config/Config.kt
```diff
@@ -26,6 +26,8 @@ data class ExecutionClientConfig(
 
 data class P2P(
   val port: UInt = 9000u,
+  val bootNodeP2PAddress: String? = null,
```
Filter94, line 29: Can it be an url?

Also, maybe we should support multiple bootnodes?

jframe: I think this will be a comma-separated list of enrs like in Teku https://docs.teku.consensys.io/reference/cli#p2p-discovery-bootnodes

@pinges Is this what you are thinking about the discovery nodes?

PR #47, opened 2025-02-27
Qbft block creation

jframe:
Implements beacon block creation and adaptors for qbft to create blocks.

Note: keeping this in draft as this builds on top of qbft block hashing for the types https://github.com/Consensys/maru/pull/42


consensus/src/main/kotlin/maru/consensus/BlockCreator.kt
```diff
@@ -0,0 +1,136 @@
...
+
+  /**
+   * Creates a new block with the given timestamp on of the parent header including the execution payload ready
+   * to be proposed as a block in the QBFT consensus.
+   */
+  fun createBlock(
+    timestamp: Long,
+    roundNumber: Int,
+    parentHeader: BeaconBlockHeader,
+  ): BeaconBlock? {
```
Filter94, line 52: Can we avoid nullability here? Would it be appropriate if we throw if `finishBlockBuilding` doesn't build anything?

Issue #33, opened 2025-02-13
Qbft hashing


jframe:
**Description** 

Create hashing for QBFT block headers and the associated QBFT adapters. This needs to also account for hashing the block with committed and for on-chain blocks.

**Acceptance Criteria**
- Create a hash for onchain blocks. This needs to exclude the commit seals and round number from the hashing
- Create a hash for committed seals. This needs to exclude the commit seals from the hashing but include the round number and other typical qbft extra data
- Hash must be enough to uniquely identify a block
- Provide an implementation or adaptor for QbftBlockHashing
- Provide an implementation or adaptor for QbftBlockInterface
- Provide an implementation or adaptor for QbftBlockCodecAdaptor

PR #53, opened 2025-02-28
qbft block hashing v2

jframe:
Create hashing for QBFT block headers and the associated QBFT adapters. This replaces previous PR https://github.com/Consensys/maru/pull/42 this got broken in github with the repo change.

fixes https://github.com/Consensys/maru/issues/33


pinges (commented):
1 nit

serialization/src/main/kotlin/maru/serialization/rlp/BeaconBlockHeaderSerializer.kt
```diff
@@ -16,12 +16,15 @@
...
 import org.apache.tuweni.bytes.Bytes
 import org.hyperledger.besu.ethereum.rlp.RLPInput
 import org.hyperledger.besu.ethereum.rlp.RLPOutput
 
 class BeaconBlockHeaderSerializer(
   private val validatorSerializer: ValidatorSerializer,
+  private val headerHashFunction: (Serializer<BeaconBlockHeader>) -> HeaderHashFunction,
 ) : RLPSerializer<BeaconBlockHeader> {
   override fun writeTo(
     value: BeaconBlockHeader,
```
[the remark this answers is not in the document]
jframe: It's a nicer name, but it does match the interface used for all serializers. And if we change this one, we probably should change all the serializers so leaning towards another PR for this change because it will be multiple files changed for consistency, though a trivial change.

serialization/src/main/kotlin/maru/serialization/rlp/BeaconBlockHeaderSerializer.kt
```diff
@@ -43,18 +48,23 @@ class BeaconBlockHeaderSerializer(
...
 
     return BeaconBlockHeader(
       number = number,
       round = round,
+      timestamp = timestamp,
       proposer = proposer,
       parentRoot = parentRoot,
       stateRoot = stateRoot,
+      bodyRoot = bodyRoot,
+      headerHashFunction.invoke(this),
```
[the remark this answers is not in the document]
Filter94: Also, this looks a bit strange to me. Here we have a serialized RLP input. In order to compute a hash of `BeaconBlockHeader` we're going to serialize it once again to RLP and hash it. Since it's already RLP-serialized, it might be better to just reuse the RLP input here.

I'm not 100% sure about this though as this will come at the cost of a higher memory footprint as the RLP bytes will have to be hanging in a closure somewhere. So it depends on whether hash() will ever be called on `BeaconBlockHeader`

core/src/main/kotlin/maru/core/HashUtil.kt
```diff
@@ -0,0 +1,54 @@
...
+import maru.serialization.Serializer
+import org.apache.tuweni.bytes.Bytes
+import org.hyperledger.besu.datatypes.Hash
+
+typealias HeaderHashFunction = (BeaconBlockHeader) -> ByteArray
+
+/**
+ * Utility class for hashing various parts of the beacon chain
+ */
+object HashUtil {
```
gauravahuja, line 27: This object creates a dependency on tuweni inside core. I remember having a comment on one of my earlier PR that we don't want that dependency, so can we move this to serializer package or any other?

consensus/src/main/kotlin/maru/consensus/qbft/adaptors/QbftBlockInterfaceAdaptor.kt
```diff
@@ -0,0 +1,41 @@
...
+
+import maru.core.BeaconBlock
+import org.hyperledger.besu.consensus.qbft.core.types.QbftBlock
+import org.hyperledger.besu.consensus.qbft.core.types.QbftBlockInterface
+import org.hyperledger.besu.consensus.qbft.core.types.QbftHashMode
+
+/**
+ * Adaptor class for QBFT block interface, this provides a way to replace the round number in a block
+ */
+class QbftBlockInterfaceAdaptor : QbftBlockInterface {
```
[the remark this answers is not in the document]
Filter94: Whatever is your decision, it's not a blocking comment. Since the change would require a whole Besu release, I suggest keeping it as is and improve later on

core/src/main/kotlin/maru/core/HashUtil.kt
Filter94: We might avoid it by parametrizing the Hash function for example

serialization/src/main/kotlin/maru/serialization/rlp/BeaconBlockHeaderSerializer.kt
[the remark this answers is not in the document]
gauravahuja: My 2 cents, I prefer keeping it generic, so that the hash function can serialize into any encoding, rlp, ssz, or anything else and then compute the hash. So we could use RLP for db storage and network but something else for computing hash.

core/src/main/kotlin/maru/core/HashUtil.kt
jframe: Had a go at making it generic by parameterising on the hashing function. Not sure I like it as it makes it more difficult to use, but I can work around that I guess. The downside is that the serialisation module is aware of Keccak, but only at the RLPSerializers level, not in the serialisers themselves.

gauravahuja (dismissed):
lgtm

core/build.gradle
```diff
@@ -17,3 +17,10 @@ plugins {
   id 'maru.kotlin-library-conventions'
   id 'java-test-fixtures'
 }
+
+dependencies {
+  implementation "org.hyperledger.besu:besu-datatypes"
```
Filter94, line 22: I didn't find where it's used. Can we remove it?

Filter94 (approved):

core/build.gradle
jframe: Removed it. Needed when I had the Hash type being used in core for hashing

Filter94 (approved):

gauravahuja (approved):

PR #54, opened 2025-03-06
Qbft block creator

jframe:
Implements beacon block creation and adaptors for qbft to create blocks.

This does not include creating blocks for round changes that will done in a seperate PR as the logic is more complex.


Filter94 (commented):
The review is incomplete. I'll review the tests later as well

consensus/src/main/kotlin/maru/consensus/ValidatorProvider.kt
```diff
@@ -0,0 +1,44 @@
...
+
+import kotlin.collections.setOf
+import maru.core.BeaconBlockHeader
+import maru.core.Validator
+
+/**
+ * Provides access to the set of validators for a given block.
+ */
+interface ValidatorProvider {
+  fun getValidatorsAtHead(): Set<Validator>
```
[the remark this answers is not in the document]
jframe: It turns out I don't need this method after all, so I am removing it from Besu and will just not implement it in Maru for now till I can delete it. There is a similar method on the FinalState getValidators though, but we can deal with that in that PR and also possibly remove it.

consensus/src/main/kotlin/maru/consensus/ValidatorProvider.kt
```diff
@@ -0,0 +1,44 @@
...
+
+  fun getValidatorsAfterBlock(header: BeaconBlockHeader): Set<Validator>
```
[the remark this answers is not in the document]
jframe: Agree. Spotted that and forgot to change it. done

consensus/src/main/kotlin/maru/consensus/ValidatorProvider.kt
```diff
@@ -0,0 +1,44 @@
...
+
+  fun getValidatorsForBlock(header: BeaconBlockHeader): Collection<Validator>
```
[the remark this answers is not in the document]
jframe: The ValidatorProvider in Besu required the whole `BeaconBlockHeader` so it has access to the extra data for votes and current validators. It will be hard to change this in Besu. But I can change the Maru `ValidatorProvider` and just use a block number here, so at least it's cleaner here.

consensus/src/main/kotlin/maru/consensus/qbft/adaptors/QbftBlockCreatorFactoryAdaptor.kt
```diff
@@ -0,0 +1,33 @@
...
+package maru.consensus.qbft.adaptors
+
+import maru.consensus.BlockCreator
+import org.hyperledger.besu.consensus.qbft.core.types.QbftBlockCreator
+import org.hyperledger.besu.consensus.qbft.core.types.QbftBlockCreatorFactory
+
+/**
+ * Adaptor to convert a [BlockCreator] to a [QbftBlockCreatorFactory].
+ */
+class QbftBlockCreatorFactoryAdaptor(
```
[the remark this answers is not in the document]
jframe: I need to pass it as an argument to the `QbftController` indirectly via the `QbftFinalState` so it will be used wherever the qbft protocol is initialised. so I think might more sense to leave it as.

consensus/src/test/kotlin/maru/consensus/qbft/adaptors/QbftBlockCodecAdaptorTest.kt
```diff
@@ -0,0 +1,46 @@
...
+    val testValue = QbftBlockAdaptor(beaconBlock)
+    val qbftBlockCodecAdaptor = QbftBlockCodecAdaptor()
+
+    val encodedData = RLP.encode { rlpOutput -> qbftBlockCodecAdaptor.writeTo(testValue, rlpOutput) }
+    val decodedValue = qbftBlockCodecAdaptor.readFrom(RLP.input(encodedData), QbftHashMode.COMMITTED_SEAL)
+    assertThat(decodedValue).isEqualTo(testValue)
+  }
+
+  @Test
+  fun `can encode and decode same value for onchain`() {
```
Filter94, line 37: But hash mode isn't actually used. Is it only to test that at least it doesn't fail?

consensus/src/test/kotlin/maru/consensus/qbft/adaptors/QbftBlockCreatorAdaptorTest.kt
```diff
@@ -0,0 +1,66 @@
...
+        12,
+        seals.map {
+          Seal(it.encodedBytes().toArray())
+        },
+      ),
+    ).thenReturn(DataGenerators.randomSealedBeaconBlock(round.toULong()))
+    val qbftBlockCreatorAdaptor = QbftBlockCreatorAdaptor(blockCreator, 1)
+
+    val sealedBlock = qbftBlockCreatorAdaptor.createSealedBlock(extraDataProvider, qbftBlock, round, seals)
+    assertThat(sealedBlock)
```
Filter94, line 64: This assertion seems to be cut off. I guess we should compare it to the expected one

consensus/src/test/kotlin/maru/consensus/qbft/adaptors/QbftBlockCodecAdaptorTest.kt
jframe: I was expected to use the onchain hash mode in the future. Since then I've removed that mode for the `QbftBlockCodec` from Besu so this test doesn't make sense anymore. Removing it.

consensus/src/test/kotlin/maru/consensus/qbft/adaptors/QbftBlockCreatorAdaptorTest.kt
jframe: fixed

consensus/src/main/kotlin/maru/consensus/ValidatorProvider.kt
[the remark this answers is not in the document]
Filter94: Now can't we replace `getValidatorsAfterBlock` methods with calls for `getValidatorsForBlock`?

consensus/src/main/kotlin/maru/consensus/BlockCreator.kt
```diff
@@ -0,0 +1,113 @@
...
+    val finalBlockHeader = stateRootBlockHeader.copy(stateRoot = stateRoot)
+
+    return BeaconBlock(finalBlockHeader, beaconBlockBody)
+  }
+
+  /**
+   * Creates a sealed block ready to be imported into the blockchain. This means including the commit seals
+   * and round number and replacing the block hash with the onchain hash function.
+   */
+  fun createSealedBlock(
```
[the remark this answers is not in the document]
jframe: I think that's ok. The `BlockCreator` will only ever be used by QBFT so move all the logic from that into one class.

consensus/src/main/kotlin/maru/consensus/ValidatorProvider.kt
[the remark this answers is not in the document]
jframe: No, I don't think we can do that, as we don't always have the block yet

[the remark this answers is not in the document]
Filter94: We can if we make the result nullable. Also, `getValidatorsAfterBlock`'s result has to be nullable if its result may be unavailable.

To avoid ambiguity, can you please explain what `getValidatorsAfterBlock` is supposed to return exactly?

consensus/src/main/kotlin/maru/consensus/qbft/adaptors/QbftBlockCreatorAdaptor.kt
```diff
@@ -0,0 +1,59 @@
...
+import org.hyperledger.besu.consensus.qbft.core.types.QbftBlock
+import org.hyperledger.besu.consensus.qbft.core.types.QbftBlockCreator
+import org.hyperledger.besu.consensus.qbft.core.types.QbftBlockHeader
+import org.hyperledger.besu.consensus.qbft.core.types.QbftExtraDataProvider
+import org.hyperledger.besu.crypto.SECPSignature
+
+/**
+ * Adaptor to convert a [BlockCreator] to a [QbftBlockCreator].
+ */
+class QbftBlockCreatorAdaptor(
```
Filter94, line 29: > In the United States, “adapter” is preferred whereas in other English-speaking countries like the UK and Australia, “adaptor” is more common

I respect, that, but should we stick to "Adapter" to align with GoF's terminology 🙂?

jframe: Yes, should use the US spelling

consensus/src/main/kotlin/maru/consensus/ValidatorProvider.kt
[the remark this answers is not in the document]
jframe: The method `getValidatorAfterBlock `returns the validators that are permitted to create the next block.

The method `getValidatorAfterBlock` is used in places such as block creation where we don't have the new block yet. So we can't just pass in the block because it doesn't exist yet to get the validators.

consensus/src/main/kotlin/maru/consensus/ValidatorProvider.kt
```diff
@@ -0,0 +1,39 @@
...
+  fun getValidatorsAfterBlock(blockNumber: ULong): Set<Validator>
+
+  fun getValidatorsForBlock(blockNumber: ULong): Set<Validator>
+}
+
+/**
+ * A [ValidatorProvider] that always returns the same [Validator] instance. This is useful for the single validator case.
+ */
+class StaticValidatorProvider(
+  private val validator: Validator,
```
gauravahuja, line 34: This could be a set of validators that we get from some config file.
```suggestion
  private val validators: Set<Validator>,
```

consensus/src/main/kotlin/maru/consensus/ValidatorProvider.kt
```diff
@@ -0,0 +1,39 @@
...
+package maru.consensus
+
+import kotlin.collections.setOf
+import maru.core.Validator
+
+/**
+ * Provides access to the set of validators for a given block.
+ */
+interface ValidatorProvider {
+  fun getValidatorsAfterBlock(blockNumber: ULong): Set<Validator>
```
gauravahuja, line 25: This can have a default implementation ?
```suggestion
  fun getValidatorsAfterBlock(blockNumber: ULong): Set<Validator> = getValidatorsForBlock(blockNumber + 1uL)
```

consensus/src/main/kotlin/maru/consensus/ValidatorProvider.kt
```diff
@@ -0,0 +1,39 @@
...
+
+  fun getValidatorsForBlock(blockNumber: ULong): Set<Validator>
+}
+
+/**
+ * A [ValidatorProvider] that always returns the same [Validator] instance. This is useful for the single validator case.
+ */
+class StaticValidatorProvider(
```
gauravahuja, line 33: When you rebase on main you can leave the `getValidatorsForBlock(block: BeaconBlockHeader)` unimplemented. I will refactor and change any usages to conform to your interface with block numbers.

jframe: Ended up just implementing it since it's trivial for the `StaticValidatorProvider`.

consensus/src/main/kotlin/maru/consensus/ValidatorProvider.kt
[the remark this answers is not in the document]
jframe: Changed to return a `SafeFuture`

gauravahuja (approved):

consensus/src/main/kotlin/maru/consensus/ValidatorProvider.kt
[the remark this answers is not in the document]
Filter94: So if we ended up with my initial proposal, why do we need the `getValidatorAfterBlock`? We can simplify this interface and make it functional. Do you see any drawbacks?

consensus/src/main/kotlin/maru/consensus/qbft/adaptors/BlockExtensions.kt
Filter94, line 1: Can we rename the package to `adapters` as well?

consensus/src/main/kotlin/maru/consensus/ValidatorProvider.kt
```diff
@@ -19,6 +19,27 @@ import maru.core.BeaconBlockHeader
 import maru.core.Validator
 import tech.pegasys.teku.infrastructure.async.SafeFuture
 
+/**
+ * Provides access to the set of validators for a given block.
+ */
 interface ValidatorProvider {
   fun getValidatorsForBlock(header: BeaconBlockHeader): SafeFuture<Set<Validator>>
```
Filter94, line 26: If there are methods to get validators by block number, why would user use method that accepts the header if they can just pass block number? I think we should remove this method now

Filter94: I saw Gaurav's comment now. I guess it will be removed later

Filter94 (dismissed):

consensus/src/main/kotlin/maru/consensus/qbft/adapters/QbftValidatorProviderAdaptor.kt
```diff
@@ -0,0 +1,42 @@
...
+import maru.consensus.ValidatorProvider
+import org.apache.tuweni.bytes.Bytes
+import org.hyperledger.besu.consensus.qbft.core.types.QbftBlockHeader
+import org.hyperledger.besu.consensus.qbft.core.types.QbftValidatorProvider
+import org.hyperledger.besu.datatypes.Address
+
+/**
+ * Adapter to convert a [ValidatorProvider] to a [QbftValidatorProvider].
+ */
+class QbftValidatorProviderAdaptor(
```
Filter94, line 27: Still an adaptor

consensus/src/main/kotlin/maru/consensus/qbft/adapters/QbftSealedBlockAdaptor.kt
Filter94, line 1: The file names still have Adaptor

jframe: They should be all gone now 🤞

Filter94 (approved):

Issue #35, opened 2025-02-13
Beacon chain block importer


jframe:
**Description**
Import beacon chain blocks onto the chain. This also involves telling the EL to import the block.

**Acceptance Criteria**
- Generate beacon state by processing the beacon block
- Can assume the block is already validated
- Store block and state in the database
- Notify the EL that the block is finalised by doing an FCU on the engine API
- Start building a new block on the EL if we are the next proposer

PR #63, opened 2025-03-20
[35] Block import logic

Filter94:
Addresses #35 
* Adds `QbftBlockImporter`
* Implements previously not implemented  methods in `JsonRpcExecutionLayerManager`
* Various collateral improvements and rearrangements


executionlayer/src/main/kotlin/maru/executionlayer/manager/JsonRpcExecutionLayerManager.kt
```diff
@@ -223,11 +168,63 @@ class JsonRpcExecutionLayerManager private constructor(
...
+          )
+        } else {
+          mapForkChoiceUpdatedResultToDomain(response)
+        }
+      }
 
-  override fun importBlock(executionPayload: ExecutionPayload): ByteArray {
-    TODO("Will implement once the chain following feature is implemented")
-  }
+  override fun importBlock(executionPayload: ExecutionPayload): SafeFuture<Unit> =
```
jframe, line 210: Better name for this would be `newPayload` as that is what is called in the engine API. Just doing `newPayload` by itself isn't a complete import we also need to do a FCU.

jframe (commented):
Only an initial quick review

executionlayer/src/main/kotlin/maru/executionlayer/manager/JsonRpcExecutionLayerManager.kt
Filter94: `Block` means something different here. El is importing a block, but that block isn't necessarily the head of the chain. But I'll rename this method for extra clarity.

storage/src/main/kotlin/maru/database/kv/KvDatabase.kt
```diff
@@ -52,12 +52,12 @@ class KvDatabase(
     }
   }
 
-  override fun getLatestBeaconState(): BeaconState? = kvStoreAccessor.get(Schema.LatestBeaconState).getOrNull()
+  override fun getLatestBeaconState(): BeaconState = kvStoreAccessor.get(Schema.LatestBeaconState).get()
 
-  override fun getBeaconState(beaconBlockRoot: ByteArray): BeaconState? =
+  override fun findBeaconState(beaconBlockRoot: ByteArray): BeaconState? =
```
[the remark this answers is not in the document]
Filter94: That might be true for the KV stores because they're essentially hash maps and every element is either there or it's not, so every operation would be a `find`. So sticking to just `get` is more straightforward from this perspective.

For `Database` interface (which you renamed to `BeaconChain` in your PR) I think we're losing the specificity of a KV store. `BeaconChain` is just a domain object that will be backed by a KV store in production and maybe by a simple hash map in the tests.

If you think find as "return something that may be missing" and get as "return something that _has_ to be there" is not intuitive for you, let me know, but that would be another argument.

consensus/src/main/kotlin/maru/consensus/ProposerSelector.kt
```diff
@@ -17,8 +17,12 @@ package maru.consensus
 
 import maru.core.BeaconBlockHeader
 import maru.core.Validator
+import org.hyperledger.besu.consensus.common.bft.ConsensusRoundIdentifier
 import tech.pegasys.teku.infrastructure.async.SafeFuture
 
-interface ProposerSelector {
-  fun getProposerForBlock(header: BeaconBlockHeader): SafeFuture<Validator>
+fun interface ProposerSelector {
+  fun getProposerForBlock(header: ConsensusRoundIdentifier): SafeFuture<Validator>
```
gauravahuja, line 24: If you look at the [Implementation](https://github.com/Consensys/maru/pull/66/files), I actually need to get the last block. I use the parentRoot in the block header for that.

If we change this to ConsensusRoundIdentifier, we will have to add a new method to the Database to get block by number/sequence.

Filter94: That's a good point. I think we'll need a method to get beacon block by its number. For example to implement methods like [this](https://ethereum.github.io/beacon-APIs/#/Beacon/getBlockHeader)

Filter94: I see this method in Jason's [PR](https://github.com/Consensys/maru/pull/67/files#diff-3f096aebe52405a3a89bf832d58acc059ca0750036bc44f4151ba8f806574d9bR70)

gauravahuja (approved):

Filter94:
@jframe I'm going to merge it, but I'd like to see your feedback as well

PR #64, opened 2025-03-21
Refactor proposerselector

gauravahuja:
- Proposer selector to use Consensus Round identifier

<del> - Add method to create a transition beacon block header for state hash calcluation purposes. It sets a default value of state root as ByteArray(0) </del>

### DIFF
--- a/consensus/src/main/kotlin/maru/consensus/BeaconBlockHeaderExtensions.kt
+++ b/consensus/src/main/kotlin/maru/consensus/BeaconBlockHeaderExtensions.kt
@@ -15,10 +15,11 @@
  */
 package maru.consensus
 
-import maru.core.Validator
+import maru.core.BeaconBlockHeader
 import org.hyperledger.besu.consensus.common.bft.ConsensusRoundIdentifier
-import tech.pegasys.teku.infrastructure.async.SafeFuture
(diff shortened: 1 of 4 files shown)


consensus/src/main/kotlin/maru/consensus/ProposerSelector.kt
```diff
@@ -15,10 +15,10 @@
...
 package maru.consensus
 
-import maru.core.BeaconBlockHeader
 import maru.core.Validator
+import org.hyperledger.besu.consensus.common.bft.ConsensusRoundIdentifier
 import tech.pegasys.teku.infrastructure.async.SafeFuture
 
 interface ProposerSelector {
-  fun getProposerForBlock(header: BeaconBlockHeader): SafeFuture<Validator>
+  fun getProposerForBlock(consensusRoundIdentifier: ConsensusRoundIdentifier): SafeFuture<Validator>
```
jframe, line 23: Be good to use the same method name as we do in `https://github.com/hyperledger/besu/blob/main/consensus/common/src/main/java/org/hyperledger/besu/consensus/common/bft/blockcreation/ProposerSelector.java` i.e. selectProposerForRound

PR #65, opened 2025-03-24
[60] Implement QbftBlockValidator

gauravahuja:
#60


consensus/src/main/kotlin/maru/consensus/qbft/adaptors/QbftBlockValidatorAdapter.kt
```diff
@@ -0,0 +1,44 @@
...
+import maru.consensus.state.StateTransition
+import maru.database.Database
+import org.hyperledger.besu.consensus.qbft.core.types.QbftBlock
+import org.hyperledger.besu.consensus.qbft.core.types.QbftBlockValidator
+
+class QbftBlockValidatorAdapter(
+  private val stateTransition: StateTransition,
+  private val database: Database,
+) : QbftBlockValidator {
+  override fun validateBlock(qbftBlock: QbftBlock?): QbftBlockValidator.ValidationResult {
```
Filter94, line 30: I don't think it's really nullable. In cases like that where nullability comes from Java, we should just enforce non-nullability if we know the input is never nullable

consensus/src/main/kotlin/maru/consensus/qbft/adaptors/QbftBlockValidatorAdapter.kt
```diff
@@ -0,0 +1,44 @@
...
+
+import com.github.michaelbull.result.Err
+import com.github.michaelbull.result.Ok
+import java.util.Optional
+import maru.consensus.state.StateTransition
+import maru.database.Database
+import org.hyperledger.besu.consensus.qbft.core.types.QbftBlock
+import org.hyperledger.besu.consensus.qbft.core.types.QbftBlockValidator
+
+class QbftBlockValidatorAdapter(
```
Filter94, line 26: Why isn't this class an instance of `BlockValidator`? My expectation from `QbftBlockValidator` would be to compose all the relevant validations so they're callable from QBFT core

consensus/src/main/kotlin/maru/consensus/qbft/adaptors/QbftBlockValidatorAdapter.kt
gauravahuja: This comes from java and I enforce non nullabilty on the first line of the method using !! operator.

Filter94: But since you're doing `!!` at the very beginning you can as well just use `QbftBlock` instead of `QbftBlock?` in the argument type

gauravahuja: I think that results in a kotlin compilation error

consensus/src/main/kotlin/maru/consensus/qbft/adaptors/QbftBlockValidatorAdapter.kt
Filter94: I still don't see why state transition would go into `QbftBlockValidatorAdapter`. IMO adapter's responsibility has to be only to adapt Maru's `BlockValidator` to Besu's `QbftBlockValidator` that's it. State transition is perfectly implementable as a `BlockValidator`. And `CompositeBlockValidator` enables flexible changes of validation rules. And I see state transition just as one of those rules, nothing special.

consensus/src/test/kotlin/maru/consensus/state/StateTransitionImplTest.kt
```diff
@@ -15,108 +15,118 @@
...
+        return SafeFuture.completedFuture(validator)
+      }
     }
-
-  private val proposerSelector =
+  private val proposerSelector: ProposerSelector =
     object : ProposerSelector {
-      override fun getProposerForBlock(header: BeaconBlockHeader): SafeFuture<Validator> =
-        SafeFuture.completedFuture(newBlockHeader.proposer)
+      override fun getProposerForBlock(header: BeaconBlockHeader): SafeFuture<Validator> {
```
Filter94, line 87: I think that this branching is makes the tests less readable and error prone. Since `ProposerSelector` is a functional interface, it takes no effort to instantiate it with the expected result as a lambda in the body of a test. It would improve test's readability as well, because reader wouldn't have to decipher which branch is expected for a given test.

consensus/src/test/kotlin/maru/consensus/state/StateTransitionImplTest.kt
```diff
@@ -15,108 +15,118 @@
...
-  private val postState =
-    BeaconState(
-      latestBeaconBlockHeader = newBlockHeader,
-      latestBeaconBlockRoot = newBlockHeader.bodyRoot,
-      validators = validators,
-    )
+  @BeforeEach
+  fun setup() {
+    beaconChain = mock()
+    doReturn(preState).`when`(beaconChain).getLatestBeaconState()
```
Filter94, line 101: You can instead use `whenever` to get rid of the backticks in Kotlin

consensus/src/test/kotlin/maru/consensus/state/StateTransitionImplTest.kt
```diff
@@ -15,108 +15,118 @@
...
-  private val validatorProvider =
+  private lateinit var beaconChain: BeaconChain
+  private val validatorProvider: ValidatorProvider =
     object : ValidatorProvider {
-      override fun getValidatorsForBlock(header: BeaconBlockHeader): SafeFuture<Set<Validator>> =
-        SafeFuture.completedFuture(validators)
-
-      override fun getValidatorsForBlock(blockNumber: ULong): SafeFuture<Set<Validator>> =
-        SafeFuture.completedFuture(validators)
+      override fun getValidatorsForBlock(blockNumber: ULong): SafeFuture<Set<Validator>> {
```
Filter94, line 75: Same point as below, except it's not yet a functional interface, but it will be. And the tests are one of the reason I push for the functional interfaces.

consensus/src/main/kotlin/maru/consensus/validation/BlockValidator.kt
```diff
@@ -72,149 +67,166 @@ class CompositeBlockValidator(
...
-            "Block number is not the next block number " +
-              "blockNumber=${block.beaconBlockHeader.number} " +
-              "nextBlockNumber=${parentBlockHeader.number + 1u}",
-          ),
-        )
-      } else {
-        SafeFuture.completedFuture(ok())
-      }
+class BlockNumberValidator(
+  private val beaconChain: BeaconChain,
```
Filter94, line 71: I noticed that `beaconChain.getLatestBeaconState()` is called in multiple validators. Wouldn't it be better to pass BeaconState into a constructor instead?

consensus/src/test/kotlin/maru/consensus/validation/BlockValidatorTest.kt
```diff
@@ -64,8 +71,61 @@ class BlockValidatorTest {
     )
   private val validNewBlock = BeaconBlock(validNewBlockHeader, validNewBlockBody)
 
+  private lateinit var beaconChain: BeaconChain
+  private lateinit var proposerSelector: ProposerSelector
+  private lateinit var validatorProvider: ValidatorProvider
+
+  @BeforeEach
+  fun setup() {
+    beaconChain = mock()
```
Filter94, line 80: I created an in memory implementation for the tests in my [PR](https://github.com/Consensys/maru/pull/63/files#diff-c54f829cd9d3a6bba7f17a3125c2cdd910751b09eb8e899592bbcf5415e9f985R21), but it's up to you if you want to use it

gauravahuja: Can you merge that, I approved it.

consensus/src/main/kotlin/maru/consensus/qbft/adaptors/QbftBlockValidatorAdapter.kt
Filter94: I didn't mean that `StateTransition` should extend `BlockValidator`. I mean that the logic that is validates state transition should be performed in a `BlockValidator`. And `QbftBlockValidatorAdapter` should just delegate to a `BlockValidator`

consensus/src/test/kotlin/maru/consensus/validation/BlockValidatorTest.kt
Filter94: I just did

consensus/src/main/kotlin/maru/consensus/qbft/adaptors/QbftBlockValidatorAdapter.kt
gauravahuja: I see state transition and block validation as two separate concepts.

In the case I create a new block validator that has to validate the state transition it will need to know how to build the new state which is not something a block validator should be doing.

Filter94: That's true, but in some cases we're executing a state transition solely for validation purposes. That's exactly what's happening in `QbftBlockValidatorAdapter` class. We're applying state transition function without actually mutating the state and its result won't be used for actual state transition during block import

consensus/src/main/kotlin/maru/consensus/validation/BlockValidator.kt
gauravahuja: If I inject the latest beacon state into the block validator, QBFT block validator adapter will have to instantiate new validators on each block. To do so it will require a block validator factory that accepts the the latest state and instantiate a composite validator with all other validators.

Keeping it as is, as you suggested, I can just do a delegate call in the QBFT block validator.

consensus/src/main/kotlin/maru/consensus/qbft/adaptors/QbftBlockValidatorAdapter.kt
gauravahuja: ok, I removed any validation from State Transition and moved it to a new StateRootValidator. Also the State Transition now assumes the block is valid and simply uses it to build new state.

consensus/src/test/kotlin/maru/consensus/state/StateTransitionImplTest.kt
gauravahuja: Removed

consensus/src/test/kotlin/maru/consensus/state/StateTransitionImplTest.kt
gauravahuja: Removed

consensus/src/main/kotlin/maru/consensus/qbft/adaptors/QbftBlockValidatorAdapter.kt
jframe: Don't have much to add here; I agree with @Filter94 that we are using the state transition to validate the state. And I also expected to see a single validator that we would be able to call from the adapter. So having a `BlockValidator` that uses the state transition to validate the state along with the other block validation would work well I think.

core/src/main/kotlin/maru/core/BeaconBlockHeader.kt
```diff
@@ -41,4 +41,8 @@ data class BeaconBlockHeader(
   override fun hashCode(): Int = hash.contentHashCode()
 
   fun hash(): ByteArray = hash
+
+  companion object {
+    val EMPTY_STATE_ROOT = ByteArray(0)
```
jframe, line 46: I think this should be `ByteArray(32)` so it is always a empty bytes32 instead of changing the size for the special case

consensus/src/main/kotlin/maru/consensus/qbft/adapters/QbftBlockValidatorAdapter.kt
```diff
@@ -0,0 +1,23 @@
...
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+ */
+package maru.consensus.qbft.adapters
+
+import maru.consensus.validation.BlockValidator
+import org.hyperledger.besu.consensus.qbft.core.types.QbftBlockValidator
+
+class QbftBlockValidatorAdapter(
```
jframe, line 21: This class as it is doesn't make a lot of sense to have could just use the `BlockValidator` passed into the constructor directly in the `QbftController` constructor.

But in the default function `validateBlock(qbftBlock: QbftBlock): QbftBlockValidator.ValidationResult` defined in `BlockValidator` would make sense here. That way `QbftBlock` is only used in the qbft package. That way the `BlockValidator` can be kept general to the BeaconBlock as it will also be used for validation in other parts of the code.

gauravahuja: In my previous commit I had the default method here but then I remembered  a previous comment from @Filter94  he suggested using delegate to a BlockValidator. But yes this class does not make much sense since BlockValidator extends QbftBlockValidator.

consensus/src/main/kotlin/maru/consensus/validation/BlockValidator.kt
Filter94: My understanding is that at least some of them will be re-instantiated, but I agree that avoiding instantiation is beneficial.

Filter94: @jframe any opinion of the # of `BeaconChain` calls vs `BlockValidator` # of instantiations?

consensus/src/main/kotlin/maru/consensus/state/StateTransition.kt
```diff
@@ -18,98 +18,42 @@ package maru.consensus.state
...
 
 class StateTransitionImpl(
-  private val blockValidator: BlockValidator,
   private val validatorProvider: ValidatorProvider,
-  private val proposerSelector: ProposerSelector,
 ) : StateTransition {
   override fun processBlock(
-    preState: BeaconState,
     block: BeaconBlock,
   ): SafeFuture<Result<BeaconState, StateTransition.StateTransitionError>> {
```
Filter94, line 47: `StateTransitionError` is never instantiated now. I think that result signature can be simplified to 
```
SafeFuture<BeaconState>
```

consensus/src/main/kotlin/maru/consensus/validation/BlockValidator.kt
```diff
@@ -19,19 +19,31 @@ import com.github.michaelbull.result.Err
...
 import maru.executionlayer.extensions.hasValidExecutionPayload
 import maru.serialization.rlp.bodyRoot
+import maru.serialization.rlp.stateRoot
 import org.hyperledger.besu.consensus.common.bft.BftHelpers
+import org.hyperledger.besu.consensus.qbft.core.types.QbftBlock
+import org.hyperledger.besu.consensus.qbft.core.types.QbftBlockValidator
 import tech.pegasys.teku.infrastructure.async.SafeFuture
 
-fun interface BlockValidator {
+interface BlockValidator : QbftBlockValidator {
```
Filter94, line 46: This is a negative change IMO for 3 reasons:
* it couples `BlockValidator` to Besu's internals. The inital approach with independent `BlockValidator` and `QbftBlockValidatorAdapter` was cleaner and it 
* Now this interface violates single responsibility principle as we have to implement both `validateBlock` and `validateBlock`. Having both is also confusing for the reader IMO
* We can't make it a functional interface now, which will make the tests that are using `BlockValidator` more verbose

I don't really see the benefit of mixing `QbftBlockValidator` and `BlockValidator` into the same interface other than removal of `QbftBlockValidatorAdapter` which would be trivial anyway

consensus/src/main/kotlin/maru/consensus/validation/BlockValidator.kt
```diff
@@ -72,149 +89,196 @@ class CompositeBlockValidator(
...
-          ),
-        )
-      } else {
-        SafeFuture.completedFuture(ok())
-      }
+class BlockNumberValidator(
+  private val beaconChain: BeaconChain,
+) : BlockValidator {
+  override fun validateBlock(newBlock: BeaconBlock): SafeFuture<Result<Unit, BlockValidationError>> {
+    val currentState = beaconChain.getLatestBeaconState()
```
jframe, line 124: The use of `beaconChain.getLatestBeaconState` here means this and the other validators can only be used for validation of the next block. Certainly the most common case. I think we are also likely to use it during syncing to validate the block we receive from peers.

The code would be clearer and also more general if we used `beaconChain.getBeaconState(newBlock.beaconBlockHeader.parentRoot)` instead.

core/src/main/kotlin/maru/core/BeaconBlockHeader.kt
gauravahuja: bytes32 is specific to the hashing algorithm we are using but for a default empty value that is only used for the state root calculation, why use a 32 byte array with 0 values instead of a an empty list?

consensus/src/main/kotlin/maru/consensus/state/StateTransition.kt
gauravahuja: Done

Filter94 (approved):

jframe (approved):

PR #66, opened 2025-03-25
[61] Implement Proposer selector

gauravahuja:
#61


consensus/src/main/kotlin/maru/consensus/ProposerSelector.kt
```diff
@@ -15,10 +15,56 @@
...
+  private val validatorProvider: ValidatorProvider,
+  private val beaconChain: BeaconChain,
+  private val config: Config,
+) : ProposerSelector {
+  data class Config(
+    val changeEachBlock: Boolean,
+    val genesisBlockNumber: ULong,
+  )
+
+  override fun getProposerForBlock(header: BeaconBlockHeader): SafeFuture<Validator> {
```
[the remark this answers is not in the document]
gauravahuja: I just updated it

consensus/src/test/kotlin/maru/consensus/ProposerSelectorImplTest.kt
```diff
@@ -0,0 +1,115 @@
...
+
+      val proposerSelector = ProposerSelectorImpl(beaconChain, validatorProvider, config)
+      val consensusRoundIdentifier = ConsensusRoundIdentifier(blockNumber.toLong(), 0)
+
+      val result = proposerSelector.getProposerForBlock(consensusRoundIdentifier).get()
+      assertThat(result in validators).isTrue()
+      prevProposer = result
+      returnedValidators.add(result)
+    }
+    assertThat(returnedValidators).isEqualTo(validators)
```
Filter94, line 90: My understanding is that thanks to `changeEachBlock` out of `totalValidators` calls with consecutive block number we're getting 5 different validators. But I'm not sure if we're just lucky that the order of `returnedValidators` matches `validators` or it's the required property of `ProposerSelectorImpl`.

My understanding is that order enforcement makes this test more brittle than necessary.

consensus/src/test/kotlin/maru/consensus/ProposerSelectorImplTest.kt
```diff
@@ -0,0 +1,115 @@
...
+
+    val result = proposerSelector.getProposerForBlock(consensusRoundIdentifier).get()
+    assertThat(result).isEqualTo(genesisValidator)
+  }
+
+  @Test
+  fun `select proposer for next block`() {
+    val config =
+      ProposerSelectorImpl.Config(
+        changeEachBlock = true,
```
Filter94, line 66: Do we need a test for `changeEachBlock = false`?

consensus/src/main/kotlin/maru/consensus/ProposerSelector.kt
```diff
@@ -17,12 +17,62 @@ package maru.consensus
...
+        )
+      Validator(proposerAddress.toArray())
+    }
+  }
+}
+
 fun BeaconBlockHeader.toConsensusRoundIdentifier(): ConsensusRoundIdentifier =
   ConsensusRoundIdentifier(this.number.toLong(), this.round.toInt())
+
+fun Validator.toAddress(): Address = Address.wrap(Bytes.wrap(address))
```
jframe, line 78: This a reasonably common conversion we are doing, would be good move this out into wherever the extension functions are declared for core types.

consensus/src/main/kotlin/maru/consensus/ProposerSelector.kt
```diff
@@ -17,12 +17,62 @@ package maru.consensus
...
+        BesuProposerSelector.selectProposerForRound(
+          consensusRoundIdentifier,
+          prevBlockProposer,
+          validators
+            .stream()
+            .map {
+              it
+                .toAddress()
+            }.toList(),
+          config.changeEachBlock,
```
jframe, line 68: Is there any need to not change every block? This flag is for the benefit of IBFT which also uses this code. In Besu QBFT always sets this to true so that we change on every block. There is no config for it in Besu for QBFT. Maru could support this but I can't think why it's needed.

consensus/src/test/kotlin/maru/consensus/ProposerSelectorImplTest.kt
gauravahuja: Yes, added those tests.

consensus/src/main/kotlin/maru/consensus/ProposerSelector.kt
gauravahuja: Moved to new file.

consensus/src/main/kotlin/maru/consensus/ProposerSelector.kt
gauravahuja: It is a config parameter and we can always fix the config to use changeEachBlock = true.

consensus/src/test/kotlin/maru/consensus/ProposerSelectorImplTest.kt
Filter94: You're right. I missed that we've got a Set on both sides

consensus/src/main/kotlin/maru/consensus/ProposerSelector.kt
gauravahuja: Set the default value in the config data class to true

Filter94: But why having it as a configurable if we don't ever plan to use a different value?

consensus/src/main/kotlin/maru/consensus/ProposerSelector.kt
```diff
@@ -15,14 +15,55 @@
...
+    }
+    val prevBlockProposer =
+      beaconChain
+        .getSealedBeaconBlock(blockNumber - 1uL)
+        ?.beaconBlock
+        ?.beaconBlockHeader
+        ?.proposer
+        ?.toAddress()
+        ?: return SafeFuture.failedFuture(
+          IllegalArgumentException("Parent block not found. parentBlockNumber=${blockNumber - 1uL}"),
```
Filter94, line 51: Nit: I believe IllegalStateException is more appropriate here

Filter94 (approved):

PR #67, opened 2025-03-25
Qbft spike

jframe:
Spike of integrating Qbft into Maru

**DO NOT MERGE**


storage/src/main/kotlin/maru/database/kv/KvDatabase.kt
```diff
@@ -60,6 +67,13 @@ class KvDatabase(
   override fun getSealedBeaconBlock(beaconBlockRoot: ByteArray): SealedBeaconBlock? =
     kvStoreAccessor.get(Schema.SealedBeaconBlockByBlockRoot, beaconBlockRoot).getOrNull()
 
+  override fun getSealedBeaconBlock(blockNumber: ULong): SealedBeaconBlock? {
+    kvStoreAccessor.get(Schema.BlockRootByBlockNumber, blockNumber).getOrNull()?.let { blockRoot ->
+      return getSealedBeaconBlock(blockRoot)
+    }
+    return null
```
Filter94, line 74: I don't think you need a separate return statement. I believe you should be able to pop it before `kvStoreAccessor.get(...`

consensus/src/test/kotlin/maru/consensus/state/StateTransitionImplTest.kt
```diff
@@ -65,11 +65,14 @@ class StateTransitionImplTest {
...
         SafeFuture.completedFuture(validators)
+
+      override fun getValidatorsForBlock(blockNumber: ULong): SafeFuture<Set<Validator>> =
+        SafeFuture.completedFuture(validators)
     }
 
   private val proposerSelector =
     object : ProposerSelector {
-      override fun getProposerForBlock(header: BeaconBlockHeader): SafeFuture<Validator> =
+      override fun selectProposerForRound(header: BeaconBlockHeader): SafeFuture<Validator> =
```
Filter94, line 75: FYI, I changed it to accept `ConsensusRoundIdentifier` instead in another PR. [here](https://github.com/Consensys/maru/pull/63#discussion_r2011676843) is the relevant discussion. 

I see you've added the required `getSealedBeaconBlock(blockNumber: ULong)` to the BeaconChain, so we're aligned here

Filter94: Now I see that you've added another `ProposerSelector` accepting `ConsensusRoundIdentifier`, so this one can be removed now

build.gradle
```diff
@@ -24,6 +24,7 @@ allprojects {
   apply plugin: 'java-library'
 
   repositories {
+    mavenLocal()
```
Filter94, line 27: Should this go into `maru.kotlin-common-minimal-conventions.gradle`?

consensus/src/main/kotlin/maru/consensus/qbft/QbftManager.kt
```diff
@@ -0,0 +1,165 @@
...
+      QbftRoundFactory(
+        finalState,
+        blockInterface,
+        protocolSchedule,
+        minedBlockObservers,
+        messageValidatorFactory,
+        messageFactory,
+      )
+
+    val transitionLogger = QbftValidatorModeTransitionLoggerAdapter()
```
Filter94, line 131: What kind of transition is it? From a non-validator to a validator and vise versa?

build.gradle
jframe: I was planning to remove this before any code is merged in. Just a temporary until we have a Besu release out with the changes I need.

consensus/src/main/kotlin/maru/consensus/qbft/QbftManager.kt
jframe: It's a transition to a new block. I don't think we need this interface; we should be able to remove it and just add a subscriber to the new block event in Besu. If I can do that, I'll remove this as I don't like this `QbftValidatorModeTransitionLogger`.

PR #68, opened 2025-03-25
Removed mergenetsplitblock from Geth and Besu genesis files. Nethermi…

Filter94:
…nd now syncs from scratch.

The issue was that `mergenetsplitblock` field in Geth and Besu genesis configs prevented Nethermind from synching, which needs `MergeForkIdTransition`. So I removed it completely, since it's not necessarily known in advance before the switch.

### DIFF
--- a/docker/compose.yaml
+++ b/docker/compose.yaml
@@ -40,7 +40,6 @@ services:
       - ./initialization/:/initialization/:ro
     ports:
       - "8545:8545"
-      - "8546:8546"
       - "8550:8550"
       - "8551:8551"
       - "30303:30303"
@@ -63,7 +62,6 @@ services:
       - ./jwt:/jwt:ro
     ports:
       - "8555:8545"
-      - "8556:8546"
       - "8561:8551"
     networks:
       - linea
@@ -88,7 +86,6 @@ services:
       - ./jwt:/jwt:ro
     ports:
       - "8565:8545"
-      - "8566:8546"
       - "8571:8551"
     networks:
       - linea
@@ -118,7 +115,6 @@ services:
       - ./initialization/:/initialization/:ro
     ports:
       - "9545:8545"
-      - "9546:8546"
       - "9550:8550"
       - "9551:8551"
     networks:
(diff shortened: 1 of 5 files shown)


docker/compose.yaml
```diff
@@ -40,7 +40,6 @@ services:
       - ./initialization/:/initialization/:ro
     ports:
       - "8545:8545"
-      - "8546:8546"
```
jframe, line 43: Is this just removing the websocket ports? Makes sense if it is, as we don't need them exposed

e2e/src/acceptanceTest/kotlin/maru/e2e/CliqueToPosTest.kt
```diff
@@ -159,13 +155,22 @@ class CliqueToPosTest {
     restartNodeFromScratch(nodeName, nodeEthereumClient)
     log.info("Container $nodeName restarted")
 
-    await
-      .pollInterval(1.seconds.toJavaDuration())
+    val awaitCondition =
+      if (nodeName.contains("nethermind")) {
+        await
+          .pollInterval(10.seconds.toJavaDuration())
```
jframe, line 161: Is nethermind slower to startup?

docker/compose.yaml
Filter94: Yes

e2e/src/acceptanceTest/kotlin/maru/e2e/CliqueToPosTest.kt
Filter94: Yes. It sometimes doesn't peer on startup and that's why sometimes it has to be restarted here. And when it's being restarted, it takes much longer to start up than Erigon for example.

The first issue with peering would be better handled by restarts by Docker compose healthchecks instead, same as Geth or Erigon, but Nethermind's JsonRPC is protected with JWT, so it's not as easy to implement the healthcheck. I decided to do it this way than spending more time to address it in docker-compose at this stage.

jframe (approved):

PR #70, opened 2025-03-28
Add getSealedBeaconBlock by number to Database and BeaconChain

gauravahuja:


storage/src/main/kotlin/maru/database/kv/KvDatabase.kt
```diff
@@ -44,6 +44,13 @@ class KvDatabase(
           KvStoreSerializers.SealedBeaconBlockSerializer,
         )
 
+      val SealedBeaconBlockByBlockNumber: KvStoreColumn<ULong, SealedBeaconBlock> =
```
jframe, line 47: Suggest changing the value to be the hash instead of the block otherwise, we are going to double the storage needed for blocks. This is what we do in Besu and then use this hash to get the block. If aggressively prune the chain it might be ok to leave it, but I think there will be a case for an archive node for the full state.

gauravahuja: Done

storage/src/main/kotlin/maru/database/kv/KvDatabase.kt
```diff
@@ -60,6 +67,15 @@ class KvDatabase(
...
 
+  override fun getSealedBeaconBlock(beaconBlockNumber: ULong): SealedBeaconBlock? =
+    kvStoreAccessor.get(Schema.BeaconBlockRootByBlockNumber, beaconBlockNumber).getOrNull().let { blockRoot ->
+      if (blockRoot != null) {
+        kvStoreAccessor.get(Schema.SealedBeaconBlockByBlockRoot, blockRoot).getOrNull()
+      } else {
+        null
+      }
+    }
+
```
Filter94, line 78:
```suggestion
  override fun getSealedBeaconBlock(beaconBlockNumber: ULong): SealedBeaconBlock? =
    kvStoreAccessor
      .get(Schema.BeaconBlockRootByBlockNumber, beaconBlockNumber)
      .flatMap { blockRoot -> kvStoreAccessor.get(Schema.SealedBeaconBlockByBlockRoot, blockRoot) }
      .getOrNull()
```

Filter94 (approved):

jframe (approved):

Filter94 (approved):

PR #71, opened 2025-03-28
Remove latestBeaconBlockRoot from BeaconState

gauravahuja:
After https://github.com/Consensys/maru/pull/70 and https://github.com/Consensys/maru/pull/65 we can remove `latestBeaconBlockRoot` from BeaconState 
- We already have body root in `latestBeaconBlockHeader` 
- We can also calculate block root using `latestBeaconBlockHeader.hash`


jframe (dismissed):
LGTM

Filter94 (approved):

Filter94 (approved):

Issue #48, opened 2025-02-27
Block creation for roundchanges


jframe:
For round changes, there is no prior block import event to start building the block, so we must immediately do a FCU with payload attributes to start building the block and use get payload to get the block.

We must ensure that we give the EL enough time to build the block otherwise, we may get an empty block. We might have to retry several times to get a non-empty block.

Filter94:
With the QBFT modifications we'd like to do, maybe we can adjust this logic. In the modified QBFT, on the round 2+ it's only possible to propose a proposal that was proposed on round 1 OR an empty block. 

## Solution with possible non empty blocks on round 2+

I identified the following modifications in the code that we'd have to do to implement the modified QBFT:

### Proposal
* In `QbftBlockHeightManager#doRoundChange` there's a logic to store the prepared certificate for the round change. It will only store it if there was a quorum on PREPARE. This is not what we need. As an option, we can change the logic in `RoundState#constructPreparedCertificate`. This is acheivable by providing an alternative implementation of `QbftRoundFactory` that would return the regular `QbftRound` with alternative `RoundState` implementation.
* In `QbftRound#startRoundWith` previously prepared block is already retrieved by roundChangeArtifacts.getBestPreparedPeer(), so it will be proposed the next round if it was ever received by the proposing validator, given the previous point is addressed.

### Block production

* `QbftBlockCreatorFactory` would have to return an alternative `QbftBlockCreator` depending on a round number. Alternative block creator would have to:
	* Import the original proposal with `ExecutionLayerManager#importPayload` if current proposer has one instead of `ExecutionLayerManager#finishBlockBuilding` in the usual flow
	* Run `ExecutionLayerManager#setHeadAndStartBlockBuilding` followed by `ExecutionLayerManager#finishBlockBuilding` if it doesn't have the proposal. This should produce an empty block on top of the previous one.
	* Execute the rest of the existing logic in `QbftBlockCreator`
* Previous point means that `QbftBlockCreatorFactory` has to be aware of the initial proposal. So `QbftRoundFactory#createNewRoundWithState` will probably have to be modified to pass the whole `RoundState` instead of the `roundNumber` to the `QbftBlockCreatorFactory#create` call.

### Validation
* New validator will be required that would deem a block with round 2+ valid in either of cases:
	* Block is empty
	* Block is non empty and there's a signed proposal for the same block with round 1 from the expected proposer. Expected proposer is derived based on the previous block header
[...]

PR #72, opened 2025-03-31
[48] - RoundChange handling logic

Filter94:
Implements #48

* added EagerBockCreator
* modified QbftBlockCreatorFactory
* moved BesuFactory into dedicated test package
* improvements around data generation
* improvements in BlockValidator
* implementation of EmptyBlockValidator


consensus/src/main/kotlin/maru/consensus/qbft/DelayedQbftBlockCreator.kt
```diff
@@ -41,15 +41,39 @@ import org.hyperledger.besu.consensus.qbft.core.types.QbftBlockHeader
...
+    fun createSealedBlock(
+      block: QbftBlock,
+      roundNumber: Int,
+      commitSeals: Collection<SECPSignature>,
+    ): QbftBlock {
+      val seals =
+        commitSeals.map {
+          Seal(it.encodedBytes().toArrayUnsafe())
+        }
+      val block1 = block.toBeaconBlock()
```
gauravahuja, line 65: It would be more readable if you rename 
block -> qbftBlock
block1 -> beaconBlock

consensus/src/main/kotlin/maru/consensus/qbft/DelayedQbftBlockCreator.kt
```diff
@@ -41,15 +41,39 @@ import org.hyperledger.besu.consensus.qbft.core.types.QbftBlockHeader
...
+      val beaconBlockHeader = block1.beaconBlockHeader
+      val updatedBlockHeader = beaconBlockHeader.copy(round = roundNumber.toUInt())
```
gauravahuja, line 67: What is the update here? Isnt `roundNumber==beaconBlockHeader.round` ?

Filter94: I think it's for the case when a block was created on a previous round and the proposal is reused without `createBlock` call.

@jframe Is this correct?

consensus/src/main/kotlin/maru/consensus/qbft/DelayedQbftBlockCreator.kt
Filter94: Done

consensus/src/main/kotlin/maru/consensus/qbft/DelayedQbftBlockCreator.kt
jframe: That's it. If we reuse the proposal from a previous round, the round number is only updated when creating the sealed block.

consensus/src/test/kotlin/maru/consensus/qbft/EagerQbftBlockCreatorTest.kt
```diff
@@ -0,0 +1,207 @@
...
+import tech.pegasys.teku.infrastructure.time.SystemTimeProvider
+
+class EagerQbftBlockCreatorTest {
+  private var cluster =
+    Cluster(
+      ClusterConfigurationBuilder().build(),
+      NetConditions(NetTransactions()),
+      ThreadBesuNodeRunner(),
+    )
+  private val besuInstance =
```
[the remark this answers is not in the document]
Filter94: True. Do you want to move it to a separate source set?

consensus/src/main/kotlin/maru/consensus/validation/BlockValidator.kt
```diff
@@ -256,12 +236,22 @@ class ExecutionPayloadValidator(
...
-        )
+      BlockValidator.require(
+        newPayloadResponse.isSuccess && newPayloadResponse.payload.hasValidExecutionPayload(),
+      ) {
+        "Execution payload validation failed: ${newPayloadResponse.errorMessage}"
       }
     }
 }
+
+object EmptyBlockValidator : BlockValidator {
```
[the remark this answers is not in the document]
Filter94: Not in the future. Immediately

consensus/src/main/kotlin/maru/consensus/qbft/DelayedQbftBlockCreator.kt
```diff
@@ -100,20 +124,5 @@ class QbftBlockCreator(
...
-    val updatedBlockHeader = beaconBlockHeader.copy(round = roundNumber.toUInt())
-    val sealedBlockBody =
-      SealedBeaconBlock(
-        BeaconBlock(updatedBlockHeader, block1.beaconBlockBody),
-        seals,
-      )
-    val sealedBeaconBlock = sealedBlockBody
-    return QbftSealedBlockAdapter(sealedBeaconBlock)
-  }
+  ): QbftBlock = Companion.createSealedBlock(qbftBlock = block, roundNumber = roundNumber, commitSeals = commitSeals)
```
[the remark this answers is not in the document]
Filter94: I removed it, thanks!

consensus/src/main/kotlin/maru/consensus/qbft/DelayedQbftBlockCreator.kt
gauravahuja: The method or parameter names do not convey the purpose, the method is updating the round number and creating a new sealed block

consensus/src/main/kotlin/maru/consensus/qbft/EagerQbftBlockCreator.kt
```diff
@@ -0,0 +1,68 @@
...
+  )
+
+  override fun createBlock(
+    headerTimeStampSeconds: Long,
+    parentHeader: QbftBlockHeader,
+  ): QbftBlock {
+    val beaconBlockHeader = parentHeader.toBeaconBlockHeader()
+    val finalizedState = finalizationStateProvider(beaconBlockHeader)
+    manager
+      .setHeadAndStartBlockBuilding(
```
gauravahuja, line 52: if we are going to reuse the block from previous round why invoke start block building?

consensus/src/main/kotlin/maru/consensus/qbft/EagerQbftBlockCreator.kt
```diff
@@ -0,0 +1,68 @@
...
+        headHash = manager.latestBlockMetadata().blockHash,
+        safeHash = finalizedState.safeBlockHash,
+        finalizedHash = finalizedState.finalizedBlockHash,
+        nextBlockTimestamp = headerTimeStampSeconds,
+        feeRecipient = blockBuilderIdentity.address,
+      ).get()
+    Thread.sleep(config.blockBuildingDuration.inWholeMilliseconds)
```
gauravahuja, line 59: I feel a better way would be to have a retryier to query EL if block is built or not instead of sleeping a fixed duration.

consensus/src/test/kotlin/maru/consensus/qbft/EagerQbftBlockCreatorTest.kt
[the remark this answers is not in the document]
jframe: Yes, moving an integration source set would be good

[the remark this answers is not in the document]
Filter94: Why? 

I deliberately want to avoid special source sets for Maru tests as much as possible. Treating "medium" tests same as unit tests brings many benefits:
* It forces them to be as independent as "small" tests, if they're not, they will get flacky and start to fail
* Related to points above, It enables parallelization of all the tests at the same time
* There is no additional configuration for `integrationTest` is needed and there is no chance of missing integration tests by using a wrong gradle command (`test` vs `check` or however we configure it)
* If an integration test gets too slow if forces us to speed it up, not hiding the discomfort

Right now we only got `acceptanceTest` as a separate source set because tests there are significantly slower, use Docker and dedicated ports that can't be reused in other tests. Until we have more tests like that, I don't the benefit in `integrationTest` source set

consensus/src/main/kotlin/maru/consensus/qbft/EagerQbftBlockCreator.kt
Filter94: I don't see how it fits here for 2 reasons:
* There's a bug in Besu, when the getPayload is called too often, Besu will just return empty block. If we'd set a retryer to 100 milliseconds, it's very likely to get an empty block
* The block building isn't "all or nothing". The longer Besu is building a block, the better block it's going to be. So we should give it at much time as possible, because we don't need "A" block, we need the best we can build in the timeframe we have.

Filter94:
> If it is creating the block, it is the proposer and should simply use its address without using the proposer selector.

The only case where it could be useful I see is when block signer doesn't match the beneficiary. But throughout the rest of the QBFT codebase these 2 roles are coupled together too tightly. So I agree we should be able to just use the validator here.

@jframe any other reason it was written like that I'm missing?

consensus/src/main/kotlin/maru/consensus/qbft/DelayedQbftBlockCreator.kt
Filter94: What is your suggestion on the method signature?

Filter94:
> I am having a hard time understanding when and on which node is the createBlock/createSealedBlock invoked?

You can refer to `QbftBlockCreator` in Besu codebase to see when it's going to be called. Since these are block production methods and called as a part of QBFT algorithm, both will only be called on Validator nodes, never on follower nodes.

consensus/src/test/kotlin/maru/consensus/qbft/EagerQbftBlockCreatorTest.kt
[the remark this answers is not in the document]
Filter94: @fluentcrafter @gauravahuja I'm curious to hear your opinion on this too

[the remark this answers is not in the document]
fluentcrafter: I would not draw the line in **Unit vs Integration**, but rather on resources/setup necessary for the test (time is a resource as well). So, if the test 1) can run inside the JVM only (no need for docker or other heavy lift resources), 2) does not depend on other external Services/APIs I would feel tempted to keep in the same test suite.

However, I can easily anticipate an exception to the above: networking/Sycnhing tests where we want to run multiple clients (Maru+Besu) in the same machine with a chain for some time (1+minutes) and misbehaving nodes, nodes entering/leaving the rollup etc. These kinds of tests are longer and more complex so would put them perhaps even in a dedicated gradle module/testsuite.

Hope this is useful.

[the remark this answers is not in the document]
Filter94: @fluentcrafter agree, that's how I think of it as well

jframe:
> > If it is creating the block, it is the proposer and should simply use its address without using the proposer selector.
> 
> The only case where it could be useful I see is when block signer doesn't match the beneficiary. But throughout the rest of the QBFT codebase these 2 roles are coupled together too tightly. So I agree we should be able to just use the validator here.
> 
> @jframe any other reason it was written like that I'm missing?

We could use our local node address here instead. Using the beneficiary won't work in all cases though as that could be changed.

consensus/src/test/kotlin/maru/consensus/qbft/EagerQbftBlockCreatorTest.kt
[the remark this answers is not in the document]
jframe: On the benefits you mentioned
- With separate source sets, you can still run tests in parallel
- Flaky tests aren't unique to integration tests. Unless people are ignoring these tests then I don't see how keeping them in the same source set helps

Integration tests are a different kind of test to the unit tests, and it is cleaner to separate those into a different source set. The dependencies and setup required for integration tests are also different to those of the unit tests. Simply distinguishing them based on the time taken to run ignores the differences between the test types.

@pinges @siladu What do you think?

consensus/src/main/kotlin/maru/consensus/qbft/EagerQbftBlockCreator.kt
Filter94: If createBlock was called on this class, it means that block from the previous round isn't reused.

jframe (approved):

consensus/src/test/kotlin/maru/consensus/qbft/EagerQbftBlockCreatorTest.kt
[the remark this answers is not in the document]
siladu: Bit late to the party here, and not directly involved in this codebase, but since he asked I agree with @jframe that it's better in general to separate, as it affords more flexibility in managing the tests, locally and in CI.

> It forces them to be as independent as "small" tests, if they're not, they will get flacky and start to fail

That should be the goal for everything except ATs anyway IMO, regardless of project structure.

> There is no additional configuration for integrationTest

Fair point, though it's a minor amount of config and a well known convention, but agree it is more moving parts.

> If an integration test gets too slow if forces us to speed it up, not hiding the discomfort

Is the implication here that the dev would skip running the integration tests entirely? 

I don't think we need to attempt to enforce behaviour through project structure for this. 
If you don't want to wait for all tests to run, there are always ways around it like running individual tests/packages.

Slowing down and introducing flaky tests can be addressed at the PR/CI stage if the author didn't spot it themselves.

Personally I think keeping integration tests separate is more flexible and conventional in my experience (therefore a more expected/less surprising project structure). The benefit of this being less friction as you move between repos.

PR #73, opened 2025-03-31
Add config to set genesis state for kvdatabase

gauravahuja:


storage/src/main/kotlin/maru/database/kv/KvDatabase.kt
```diff
@@ -59,7 +65,10 @@ class KvDatabase(
     }
   }
 
-  override fun getLatestBeaconState(): BeaconState = kvStoreAccessor.get(Schema.LatestBeaconState).get()
+  override fun getLatestBeaconState(): BeaconState =
+    kvStoreAccessor
+      .get(Schema.LatestBeaconState)
+      .getOrDefault(config.genesisState)
```
Filter94, line 71: This is not going to work with parent header validations, we need to actually put some initial state into DB

Filter94:
I implemented DB initialization [here](https://github.com/Consensys/maru/pull/78/files#diff-51489b4db80fdd6c5d2f955d234572de78ee3d17772a1c68800c0b16d1db5d24)

PR #91, opened 2025-05-05
[88] Fixed flackyness in EagerQbftBlockCreator test

### LINKED ISSUE #88: Flacky test in EagerQbftBlockCreatorTest
## Impact
__environment [CI]__

__an user is impacted [Y], if yes type of user: [Internal]__

## Describe the bug
The test `EagerQbftBlockCreatorTest > can create a non empty block with new timestamp` is flacky, [run](https://github.com/Consensys/maru/actions/runs/14748501199/job/41400386230)

### Steps to Reproduce
No consistent way to reproduce it

Filter94:
(Hopefully) Fixes #88

### DIFF
--- a/consensus/src/test/kotlin/maru/consensus/qbft/EagerQbftBlockCreatorTest.kt
+++ b/consensus/src/test/kotlin/maru/consensus/qbft/EagerQbftBlockCreatorTest.kt
@@ -17,6 +17,8 @@ package maru.consensus.qbft
 
 import java.time.Clock
 import java.time.Duration
+import java.time.Instant
+import java.time.temporal.ChronoUnit
 import kotlin.random.Random
 import kotlin.time.Duration.Companion.milliseconds
 import maru.consensus.ConsensusConfig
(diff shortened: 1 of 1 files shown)


consensus/src/test/kotlin/maru/consensus/qbft/EagerQbftBlockCreatorTest.kt
```diff
@@ -225,8 +312,35 @@ class EagerQbftBlockCreatorTest {
...
+    executionLayerManager.setHeadAndStartBlockBuilding(
+      headHash = genesisBlockHash,
+      safeHash = genesisBlockHash,
+      finalizedHash = genesisBlockHash,
+      nextBlockTimestamp = rejectedBlockTimestamp,
+      feeRecipient = validator.address,
+    )
+    val transaction = TransactionsHelper().createTransfers(1u)
+    besuInstance.execute(transaction)
+    Thread.sleep(1000)
```
gauravahuja, line 335: Why is this sleep needed?

gauravahuja (approved):

consensus/src/test/kotlin/maru/consensus/qbft/EagerQbftBlockCreatorTest.kt
```diff
@@ -99,29 +93,43 @@ class EagerQbftBlockCreatorTest {
...
-        .toDomain()
-    val parentBlock =
-      BeaconBlock(
-        beaconBlockHeader = DataGenerators.randomBeaconBlockHeader(0U),
-        beaconBlockBody =
-          DataGenerators
-            .randomBeaconBlockBody()
-            .copy(executionPayload = genesisExecutionPayload),
+  @BeforeEach
+  fun beforeEach() {
```
gauravahuja, line 98: Doesn't instantiating these variables before each test limit how the tests can be run in parallel. Since these are shared variables, this test suite cannot run multiple tests in separate threads concurrently?

consensus/src/test/kotlin/maru/consensus/qbft/EagerQbftBlockCreatorTest.kt
Filter94: To give Besu time to build a block. If we call `mainBlockCreator.createBlock` right away, it will build an empty block, which is as good as not trying to build any blocks before the test at all

consensus/src/test/kotlin/maru/consensus/qbft/EagerQbftBlockCreatorTest.kt
Filter94: That's generally a very good point, but this isn't a pressing issue ATM. Our current target is 8 minutes for the CI run and we're well within that goal. I won't spend more time on it in this PR, but I'll try to keep it in mind for the future.

Filter94: After a more careful look what you say is true only if we enable 1 test instance per multiple threads, which is not the default https://www.baeldung.com/junit-testinstance-annotation#test-instance. By Default it's 1 instance per test and I don't really see a reason why would we want 1 instance per process, since we don't manage any expensive resources in memory

Issue #87, opened 2025-04-30
Fix Delayed block creator in QBFT


jframe:
## Description
The delayed block creator is currently broken and always uses round 1.

## Motivation
Using the delayed block creator would have more time to build the block. As the block building process starts as soon as a new block is imported,

PR #100, opened 2025-05-19
Use the delayed block creator for creating blocks

jframe:
Fixes the delayed block creator so this can be used in Maru again.

fixes https://github.com/Consensys/maru/issues/87

changes
* Qbft now uses the Delayed block creator for the first round. For the first block after startup and for rounds > 0 the eager block creator is used
* The protocol starter now waits until the next timestamps is reached before starting the new consensus mechanism. In the end to end tests we were seeing this started earlier than it should of been causing issues with transactions being included.
* Eager block creator now uses a minimum block creation time instead of the entire time until the next schedule block. This is done so there is larger time gap between when block is imported and we start building the new block. Using all the available time means there no time to build a block for the delayed block creator following the eager block creator. In the tests this would cause an empty block to be created.
* Modified config for Besu so we rebuilding the block more frequently for tests. In the tests we have a very narrow time to get the transaction sent before the new block is created. Reducing the time needed between block rebuilds makes it easier for the transaction to get included for tests.
* Fixed issue with gas price used by the e2e tests. The gas price was set the same the needed gas price. Whilst valid, it means there is no priority fee and the values this block the same as an empty block. So order to not select empty blocks we need the transactions to have some value. 
* Disabled parallelism in the MaruQbftTest as this was causing instability in the test


app/src/test/kotlin/maru/app/MaruQbftTest.kt
```diff
@@ -48,7 +48,9 @@ import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.Disabled
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.io.TempDir
+import org.junit.jupiter.api.parallel.Isolated
 
+@Isolated
```
Filter94, line 53: While this brings value short term to make the CI more stable, I think we should address it anyway and maybe create an issue to make it parallel once we have time

core/src/main/kotlin/maru/executionlayer/manager/ExecutionLayerManager.kt
```diff
@@ -88,11 +88,15 @@ interface ExecutionLayerManager {
 
   fun finishBlockBuilding(): SafeFuture<ExecutionPayload>
 
+  fun hasStartedBlockBuilding(): Boolean
```
Filter94, line 91: I believe you can't guarantee its consistency since the underlying node could restart any time. And I think the guarantees we can provide are as good as just assume that block building has started

consensus/src/main/kotlin/maru/consensus/qbft/QbftBlockCreatorFactory.kt
```diff
@@ -50,15 +50,20 @@ class QbftBlockCreatorFactory(
...
-      delegate = delayedQbftBlockCreator,
-      finalizationStateProvider = finalizationStateProvider,
-      blockBuilderIdentity = blockBuilderIdentity,
-      beaconChain = beaconChain,
-      nextBlockTimestampProvider = nextBlockTimestampProvider,
-      config = eagerQbftBlockCreatorConfig,
-      clock = clock,
-    )
+    val number = beaconChain.getLatestBeaconState().latestBeaconBlockHeader.number + 1u
+    return if (round == 0 && manager.hasStartedBlockBuilding()) {
```
Filter94, line 54: Alternative idea is to create `EagerQbftBlockCreator` only on the first call of this class without the use of `hasStartedBlockBuilding`. I'm not 100% sure if it will pass the tests though. Related to [this](https://github.com/Consensys/maru/pull/100/files#r2101917455) comment

app/src/test/kotlin/maru/testutils/MaruFactory.kt
```diff
@@ -49,7 +49,7 @@ object MaruFactory {
     endpoint = "$ethereumJsonRpcUrl"
 
     [qbft-options]
-    communication-margin=200m
+    communication-margin=10m
```
Filter94, line 52: I believe that's not enough to send and process requests. Do we need this delay at all?

config/src/test/kotlin/maru/config/HopliteFriendlinessTest.kt
```diff
@@ -166,6 +166,7 @@ class HopliteFriendlinessTest {
   private val qbftOptions =
     """
     communication-margin=100m
+    min-block-build-time=1000m
```
Filter94, line 169: Should we instead pull it from the protocol's block time?

jframe: no, it should be a smaller so that there is sufficient time to start building the next block for the delayed block creator for the next block. I just chose a number for testing the config that was the same as the block time by coincidence.

consensus/src/main/kotlin/maru/consensus/qbft/QbftBlockCreatorFactory.kt
jframe: I like that idea, going to give it ago. If it works out it will be simpler and I don't need to change the `ExecutionLayerManager` then too.

core/src/main/kotlin/maru/executionlayer/manager/ExecutionLayerManager.kt
jframe: True, will try that use delayed block creator on first use approach. If that works I can remove this.

config/src/test/kotlin/maru/config/HopliteFriendlinessTest.kt
Filter94: Then isn't it equivalent to blockTime - communicationMargin which we already had before your changes?

app/src/test/kotlin/maru/app/MaruQbftTest.kt
Filter94: Please use issue templates we have in the future

consensus/src/test/kotlin/maru/consensus/qbft/QbftBlockCreatorFactoryTest.kt
```diff
@@ -0,0 +1,104 @@
...
+import maru.executionlayer.manager.ExecutionLayerManager
+import org.assertj.core.api.Assertions.assertThat
+import org.hyperledger.besu.consensus.common.bft.blockcreation.ProposerSelector
+import org.junit.jupiter.api.Test
+import org.mockito.Mockito
+import org.mockito.kotlin.whenever
+
+class QbftBlockCreatorFactoryTest {
+  private val executionLayerManager = Mockito.mock(ExecutionLayerManager::class.java)
+  private val proposerSelector = Mockito.mock(ProposerSelector::class.java)
```
Filter94, line 32: Minor: You can use Kotlin style API as we use in other places:
```
mock<ProposerSelector>
````

File: consensus/src/main/kotlin/maru/consensus/ProtocolStarter.kt [hunk 7]
```diff
@@ -76,6 +77,12 @@ class ProtocolStarter(
         newProtocolWithFork,
       )
       currentProtocolWithFork?.protocol?.stop()
+
+      // Wait until timestamp is reached before starting new protocol
+      val timeTillFork = max((nextBlockTimestamp * 1000L) - System.currentTimeMillis(), 0)
```
Filter94, line 82: Minor: I think we can factor it into a separate function for readability

Filter94 (approved):

PR #105, opened 2025-05-20
Removing EIP-2935 contract for the time being

Filter94:


docker/initialization/genesis-besu.json.template
```diff
@@ -35,11 +35,6 @@
   "baseFeePerGas": "0x7",
   "@_WARNING": "FOR LOCAL DEV ONLY - DO NOT REUSE THESE KEYS ELSEWHERE",
   "alloc": {
-    "0x0000F90827F1C53a10cb7A02335B175320002935": {
```
jframe, line 38: Why do you want to remove this? Leaving it in has no downside. And removing it will create some error logs in Besu.

Filter94: I think you're right, we better leave it

PR #106, opened 2025-05-21
Fix MaruQbftTest for latest Besu

jframe:
The MaruQbftTest was failing after the update to the latest Besu

- Added withdrawalRequestContractAddress config as this now required in Besu
- Use empty storage for withdrawals request contract so that there are no withdrawal requests returned by the contract


pinges (approved):
LGTM

PR #107, opened 2025-05-21
fix flaky test

pinges:
Fix the reconnect test. When a static peer is disconnected Maru tries to reconnect after 5 seconds. Increase the time we wait from 5.1 s to 6 seconds to make sure that the test succeeds.


p2p/src/test/kotlin/maru/p2p/P2PTest.kt
```diff
@@ -148,8 +148,9 @@ class P2PTest {
       p2pNetwork1
         .getPeer(LibP2PNodeId(PeerId.fromBase58(PEER_ID_NODE_2)))
         .get()
-        .disconnectCleanly(DisconnectReason.TOO_MANY_PEERS)
-        .thenPeek { assertNetworkHasPeers(p2pNetwork1, 0) }
+        .disconnectImmediately(Optional.empty(), true)
```
jframe, line 151: Why is this better than using the disconnect cleanly? Thought the clean disconnect would be better

pinges: The problem with the clean one is that it might take longer, and the time for the await starts ticking right after the call to close the connection. That is why I use the immediate now and await the 0 connections after.

pinges: That is even better!! Done!

jframe (approved):

Filter94 (approved):

Issue #74, opened 2025-04-02
Maru docker image


gauravahuja:
## Description
We need to complete the Dockerfile we have for Maru and add a test that would make sure that built image is at least runs and doing something. The coverage shouldn't be as extensive as E2E tests, it should be more of a smoke test

## Motivation
It would help us making sure that the image we're building is at least somewhat working

## Tasks
- [ ] Complete a Dockerfile so it builds a working Maru instance
- [ ] Create a testsuite that is using image built by Dockerfile from the previous step
- [ ] Leverage path filters and docker cache to efficiently build and reuse the existing image if the rebuild is not required
- [ ] The new testsuite should be a part of the PR's merge criteria, but it shouldn't block the merge in case no changes were made to Maru

## Acceptance criteria
- [ ] Maru image is being built on CI if there are relevant changes in the code
- [ ] The built image is being tested on CI
- [ ] The image isn't build in case it's not needed

## Risks
- [ ] Stick to best practices for the Dockerfile, keeping security in mind

## Remember to
 - [ ] Add the `documentation` label in case there is an impact on the documentation
 - [ ] Add `priority` and `team` labels
 - [ ] Add Task for updating the Runbook or adding/updating existing metrics and alerts.

jonesho:
@gauravahuja @Filter94 As a first step, can I get more description on this issue? As a newbie, just by looking at the title I have no idea what's the requirement of this issue

gauravahuja:
https://github.com/Consensys/maru/pull/49

PR #116, opened 2025-05-23
feat: support using maru container in e2e tests

jonesho:
This PR implements #74 for running the "unit and e2e tests" and "docker container smoke test" in parallel and if the workflow is "triggered manually" or from "push request to main ", the image will be pushed right after the smoke test was passed (skipped the check of "unit and e2e tests")


Filter94:
I don't think this solution serves the purpose. What we want to cover is:
* E2E test
* Maru image smoke test

We don't want to run all the E2E tests with Maru in Docker, because it's less efficient on CI and it's harder to debug locally.

For the image tests we don't really need the same coverage as we have for E2E tests and even if we have this option, we don't want to run the same tests for Maru in Docker and Maru in process on CI for economical reasons.

I think just a check if the container  doesn't fail with given config + assertion on container's logs should be sufficient

jframe:
Is PR to this fix this issue https://github.com/Consensys/maru/issues/74?

jonesho:
> @jonesho Is PR to this fix this issue #74?

@jframe Yes, and I'm working on the comment from Roman now

e2e/src/acceptanceTest/kotlin/maru/e2e/CliqueToPosTest.kt
```diff
@@ -81,24 +90,30 @@ class CliqueToPosTest {
...
+      log.info("Using Maru as docker container: $useMaruContainer")
       deleteGenesisFiles()
       dataDir.deleteRecursively()
+      dataDir.mkdirs()
       qbftCluster.before()
-
       pragueSwitchTimestamp = parsePragueSwitchTimestamp()
-      dataDir.mkdirs()
-      maru = MaruFactory.buildTestMaru(pragueSwitchTimestamp)
+      if (!useMaruContainer) {
```
Filter94, line 114: Minor: a slightly cleaner way to do it would be to define a parent abstract class with the tests themselves and have 2 separate classes. Only one of which would be active based on `useMaruContainer`

It's a quite complex refactoring though and I don't think it provides enough value ATM, so I'm fine with the way things are

.github/workflows/main.yml
```diff
@@ -48,21 +48,34 @@ jobs:
...
-      cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
     uses: ./.github/workflows/e2e-tests.yml
     with:
       e2e-tests-logs-dump: true
       has-changes-requiring-build: ${{ needs.filter-commit-changes.outputs.has-changes-requiring-build }}
     secrets: inherit
 
+  build-for-testing:
+    #needs: [ testing, run-e2e-tests ]
+    #if: ${{ always() && !cancelled() && (needs.testing.result == 'skipped' || needs.testing.result == 'success') && (needs.run-e2e-tests.result == 'skipped' || needs.run-e2e-tests.outputs.tests_outcome == 'success') }}
```
Filter94, line 59: Can we remove these comments?

Filter94 (dismissed):

PR #126, opened 2025-05-30
Discovery Service

pinges:
Adds discovery v5 to find other nodes


config/src/test/kotlin/maru/config/HopliteFriendlinessTest.kt
```diff
@@ -67,6 +67,7 @@ class HopliteFriendlinessTest {
               staticPeers = emptyList(),
               reconnectDelay = 500.milliseconds,
             ),
+          p2pConfig = P2P(ipAddress = "127.0.0.1", port = "3322", discoveryPort = "9876", staticPeers = emptyList()),
```
Filter94, line 70: Can you please add bootnodes setting somewhere here? Because right now it's not tested at all

p2p/src/main/kotlin/maru/p2p/discovery/DiscoveryService.kt
```diff
@@ -0,0 +1,187 @@
...
+
+  private val log: Logger = LogManager.getLogger(this::javaClass)
+
+  private lateinit var discoverySystem: DiscoverySystem
+
+  private val privateKey = SecretKeyParser.fromLibP2pPrivKey(Bytes.wrap(privateKeyBytes))
+  private val localNodeRecord = localNodeRecord(privateKey = privateKey)
+
+  val delayedExecutor =
+    ScheduledExecutorAsyncRunner.create(
```
Filter94, line 57: Minor: Please use commented named arguments for long lists as such

p2p/src/main/kotlin/maru/p2p/discovery/DiscoveryService.kt
```diff
@@ -0,0 +1,187 @@
...
+      "DiscoveryService",
+      1,
+      1,
+      5,
+      MetricTrackingExecutorFactory(NoOpMetricsSystem()),
+    )
+  var bootnodeRefreshTask: Cancellable? = null
```
Filter94, line 64: Should it be cancelled on `stop()`?

Filter94: Also, is `null` ever useful here as a value? Maybe we can initialize it with a dummy implementation like 
```
object : Cancellable {
      override fun cancel() {}

      override fun isCancelled(): Boolean = true
    }
```

p2p/src/main/kotlin/maru/p2p/discovery/DiscoveryService.kt
```diff
@@ -0,0 +1,187 @@
...
+        .stream()
+        .map { NodeRecordFactory.DEFAULT.fromEnr(it) }
+        .toList()
+
+    discoveryNetworkBuilder.listen(p2pConfig.ipAddress, p2pConfig.discoveryPort.toInt())
+    discoveryNetworkBuilder.secretKey(privateKey)
+    discoveryNetworkBuilder.localNodeRecord(localNodeRecord)
+    discoveryNetworkBuilder.bootnodes(bootnodes)
+
+    discoverySystem = discoveryNetworkBuilder.build()
```
Filter94, line 80: Can we build it during `DiscoveryService` construction? This way `discoverySystem` won't be a var, which is good and in case there are configuration errors, we'll fail faster

p2p/src/main/kotlin/maru/p2p/discovery/DiscoveryService.kt
```diff
@@ -0,0 +1,187 @@
...
+    discoverySystem
+      .start()
+      .thenRun {
+        discoverySystem.updateCustomFieldValue(MaruForkId.MARU_FORK_ID_FIELD_NAME, maruForkId.encode())
+        // TODO: do we want another custom field to identify topics/role/something else?
+        this.bootnodeRefreshTask =
+          delayedExecutor.runWithFixedDelay(
+            { this.pingBootnodes(bootnodes) },
+            BOOTNODE_REFRESH_DELAY,
+            { error: Throwable? ->
```
Filter94, line 93:
```suggestion
            { error: Throwable ->
```

p2p/src/main/kotlin/maru/p2p/discovery/DiscoveryService.kt
```diff
@@ -0,0 +1,187 @@
...
+            },
+          )
+      }
+  }
+
+  fun stop() {
+    discoverySystem.stop()
+  }
+
+  fun searchForPeers(): SafeFuture<Collection<MaruDiscoveryPeer>> =
```
Filter94, line 107: Does it have to be public? When should it be called by Maru?

p2p/src/main/kotlin/maru/p2p/discovery/DiscoveryService.kt
```diff
@@ -0,0 +1,187 @@
...
+    SafeFuture
+      .of(discoverySystem.searchForNewPeers())
+      // Current version of discovery doesn't return the found peers but next version will
+      .thenApply { getKnownPeers() }
+
+  fun getKnownPeers(): Collection<MaruDiscoveryPeer> =
```
Filter94, line 113: How will it be used? Should we make it package private instead, if it's supposed to be used inside P2P package only? Same for other public methods

p2p/src/main/kotlin/maru/p2p/discovery/DiscoveryService.kt
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  data class MaruDiscoveryPeer(
+    val publicKey: Bytes,
+    val nodeId: Bytes,
+    val address: InetSocketAddress,
+    val maruForkId: MaruForkId,
+  )
+
+  class MaruForkId(
```
Filter94, line 163: 1. This looks like a data object
2. I think it has to be defined somewhere in the core and be used in P2P communication as well as a part of topic id

p2p/src/test/kotlin/maru/p2p/DiscoveryTest.kt
```diff
@@ -0,0 +1,175 @@
...
+import org.awaitility.Awaitility
+import org.ethereum.beacon.discovery.schema.IdentitySchemaInterpreter
+import org.ethereum.beacon.discovery.schema.NodeRecordBuilder
+import org.ethereum.beacon.discovery.schema.NodeRecordFactory
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.parallel.Execution
+import org.junit.jupiter.api.parallel.ExecutionMode
+
+@Execution(ExecutionMode.SAME_THREAD)
+class DiscoveryTest {
```
Filter94, line 35: Can you make it run concurrently right away?

p2p/src/test/kotlin/maru/p2p/DiscoveryTest.kt
```diff
@@ -0,0 +1,175 @@
...
+        .untilAsserted {
+          val get = bootnode.searchForPeers().get()
+          assertThat(
+            get
+              .stream()
+              .filter { it.nodeId == discoveryService2.getLocalNodeRecord().nodeId }
+              .count(),
+          ).isGreaterThan(0L)
+        }
+      Awaitility
```
Filter94, line 139: I feel like this wait can be factored in a separate function for better readability and code reuse

p2p/src/test/kotlin/maru/p2p/DiscoveryTest.kt
```diff
@@ -0,0 +1,175 @@
...
+          val get = bootnode.searchForPeers().get()
+          assertThat(
+            get
+              .stream()
+              .filter { it.nodeId == discoveryService3.getLocalNodeRecord().nodeId }
+              .count(),
+          ).isGreaterThan(0L)
+        }
+    } finally {
+      bootnode.stop()
```
Filter94, line 152: Maybe it's better manage resources in `@BeforeEach` and  `@AfterEach`?

config/src/main/kotlin/maru.config/Config.kt
```diff
@@ -38,29 +38,11 @@ data class FollowersConfig(
 data class P2P(
   val ipAddress: String,
   val port: String,
+  val discoveryPort: String,
```
Filter94, line 41: I think that discovery has to be toggleable so that users enable/disable it at will. Also, it would make sense to disable it in all the current MaruApp and other tests, and enable it for the new tests only

config/src/test/kotlin/maru/config/HopliteFriendlinessTest.kt
pinges: Added a bootnode

p2p/src/main/kotlin/maru/p2p/discovery/DiscoveryService.kt
pinges: this is a java method

p2p/src/main/kotlin/maru/p2p/discovery/DiscoveryService.kt
pinges: Done

p2p/src/main/kotlin/maru/p2p/discovery/DiscoveryService.kt
pinges: Done

p2p/src/main/kotlin/maru/p2p/discovery/DiscoveryService.kt
pinges: It is package private, isn't it? 
Called by the MaruPeerManager

p2p/src/main/kotlin/maru/p2p/discovery/DiscoveryService.kt
pinges: Yes! I will use the right one now :-)

p2p/src/test/kotlin/maru/p2p/DiscoveryTest.kt
pinges: Done

config/src/main/kotlin/maru.config/Config.kt
pinges: done

fluentcrafter:
Hi @pinges, what is the status/progress on this?

core/src/main/kotlin/maru/consensus/ForkId.kt
```diff
@@ -8,64 +8,48 @@
...
-  private val beaconChain: BeaconChain,
-  private val forksSchedule: ForksSchedule,
-  private val forkIdHasher: ForkIdHasher,
-) {
-  fun currentForkIdHash(): ByteArray {
-    val forkId =
-      ForkId(
-        chainId = chainId,
-        forkSpec =
-          forksSchedule.getForkByTimestamp(
```
Filter94, line 59: While having many abstractions vs putting all the Fork Id computation logic into a single function can be debateable, the substantial difference I see here, is that with your changes fork spec is not participating in the checksum computation. This means that nodes with with different protocol rules (block time, fork switch timestamp, initial validator set) may peer to each other. I wanted to avoid that

p2p/src/main/kotlin/maru/p2p/MaruPeerManagerService.kt
```diff
@@ -0,0 +1,133 @@
...
+      discoveryService?.let { discoveryService ->
+        discoveryService
+          .searchForPeers()
+          .orTimeout(Duration.ofSeconds(30L))
+          .whenComplete { availablePeers, throwable ->
+            log.debug("Finished searching for peers. Found {} peers.", availablePeers.size)
+            if (!stopCalled) {
+              if (throwable != null) {
+                log.debug("Failed to discover peers: {}\n{}", throwable.message, throwable.stackTraceToString())
+                discoveryService.getKnownPeers().forEach { peer ->
```
Filter94, line 65: Why are we trying to reconnect to known peers when there's an exception?

p2p/src/main/kotlin/maru/p2p/MaruPeerManagerService.kt
```diff
@@ -0,0 +1,133 @@
...
+  private fun searchForPeersUntilMaxReached() {
+    if (!stopCalled && currentlySearching.compareAndSet(false, true)) {
+      discoveryService?.let { discoveryService ->
+        discoveryService
+          .searchForPeers()
+          .orTimeout(Duration.ofSeconds(30L))
+          .whenComplete { availablePeers, throwable ->
+            log.debug("Finished searching for peers. Found {} peers.", availablePeers.size)
+            if (!stopCalled) {
+              if (throwable != null) {
```
Filter94, line 63: I believe this flow can be simplified by returning the list of peers from the `if` and then unconditionally calling `tryToConnectIfNotFull(peer)`. So only the list is conditional in this case

p2p/src/main/kotlin/maru/p2p/MaruPeerManagerService.kt
```diff
@@ -0,0 +1,133 @@
...
+                availablePeers.forEach { peer ->
+                  tryToConnectIfNotFull(peer)
+                }
+              }
+            }
+          }.whenComplete { _, _ ->
+            currentlySearching.set(false)
+            log.debug("Peer count: {}. Max Peers: {}", p2pNetwork.peerCount, maxPeers)
+            if (!stopCalled && p2pNetwork.peerCount < maxPeers) {
+              CompletableFuture.runAsync { searchForPeersUntilMaxReached() }
```
Filter94, line 78: Should we delay the next call? How heavy is peer discovery?

Filter94: Also, I'm not entirely confident, but this way of running an infinite loop might have a memory leak. Consider doing it the way it's done in [LineaFinalizationProvider](https://github.com/Consensys/maru/blob/main/linea-finalization-provider/src/main/kotlin/maru/finalization/LineaFinalizationProvider.kt#L26)

p2p/src/main/kotlin/maru/p2p/MaruPeerManagerService.kt
```diff
@@ -0,0 +1,133 @@
...
+      }
+
+      log.debug("Peer {} Connecting to peer {}...", discoveryService?.getLocalNodeRecord()?.nodeId, peer.nodeIdBytes)
+      p2pNetwork
+        .connect(p2pNetwork.createPeerAddress(peer))
+        .orTimeout(30, TimeUnit.SECONDS)
+        .whenComplete { _, throwable ->
+          try {
+            if (throwable != null) {
+              log.debug("(1)Failed to connect to peer {}: {}", peer.nodeIdBytes, throwable.stackTraceToString())
```
Filter94, line 118: Do we want to leave these (1) and (2) in the logs?

p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryPeer.kt
```diff
@@ -0,0 +1,23 @@
...
+ */
+package maru.p2p.discovery
+
+import java.net.InetSocketAddress
+import java.util.Optional
+import org.apache.tuweni.bytes.Bytes
+import tech.pegasys.teku.networking.p2p.discovery.DiscoveryPeer
+
+class MaruDiscoveryPeer(
+  publicKeyBytes: Bytes?,
```
Filter94, line 17: Can we stick to ByteArray instead? Also, why is everything optional?

p2p/src/main/kotlin/maru/p2p/MaruPeerManagerService.kt
```diff
@@ -0,0 +1,133 @@
...
+        discoveryService
+          .searchForPeers()
+          .orTimeout(Duration.ofSeconds(30L))
+          .whenComplete { availablePeers, throwable ->
+            log.debug("Finished searching for peers. Found {} peers.", availablePeers.size)
+            if (!stopCalled) {
+              if (throwable != null) {
+                log.debug("Failed to discover peers: {}\n{}", throwable.message, throwable.stackTraceToString())
+                discoveryService.getKnownPeers().forEach { peer ->
+                  tryToConnectIfNotFull(peer)
```
Filter94, line 66: Since `tryToConnectIfNotFull` are creating futures that modify  `connectionInProgress` concurrently, I think just `mutableListOf` might be not enough, because it's not thread-safe

Filter94 (commented):
The review is incomplete, but I left some comments

core/src/main/kotlin/maru/consensus/ForkId.kt
```diff
@@ -8,64 +8,48 @@
...
+import java.util.zip.CRC32
+import org.apache.tuweni.bytes.Bytes
 
 data class ForkId(
   val chainId: UInt,
-  val forkSpec: ForkSpec,
   val genesisRootHash: ByteArray,
 ) {
+  companion object {
+    const val FORK_ID_FIELD_NAME = "mfid"
```
jframe, line 20: can just move this to the top of the `ForkId.kt` file outside of the data class to avoid the need for a companion object

config/src/main/kotlin/maru/config/Config.kt
```diff
@@ -39,29 +39,13 @@ data class FollowersConfig(
 data class P2P(
   val ipAddress: String,
   val port: UInt,
+  val discoveryEnabled: Boolean = true,
+  val discoveryPort: UInt,
+  val bootnodes: List<String> = emptyList(),
```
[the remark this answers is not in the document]
pinges: Bootnodes are used in discovery. as a way to find your first peer that can act as a source for other peers to discover. 
Static peers are on the p2p level. These are peers that you don't want to loose, you want to stay connected to.

jframe (commented):
Initial review comments

p2p/src/main/kotlin/maru/p2p/P2PNetworkImpl.kt
```diff
@@ -76,12 +79,35 @@ class P2PNetworkImpl(
...
+          ForkId(
+            chainId = chainId,
+            genesisRootHash = ByteArray(32),
+          )
+        },
+      )
+    } else {
+      null
+    }
+  // TODO: We need to call the updateForkId method on the discovery service when the forkId changes
```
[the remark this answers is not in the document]
pinges: In Maru ForkId is used in discovery and in the StatusMessage. I think that once we have merged one of the PRs first and then finish the usage of the ForkId. 
I think that we should always use the ForkId of out current fork, and when we hard fork to the next one start using the new one.
In Besu it is a bit more complicated, as we need the full list of ForkIds to establish whether we want to connect to a peer. I don't think that is necessary in Maru. But we can discuss that.

p2p/src/main/kotlin/maru/p2p/P2PNetworkImpl.kt
```diff
@@ -105,6 +131,10 @@ class P2PNetworkImpl(
             .createPeerAddress(peer)
             ?.let { address -> addStaticPeer(address as MultiaddrPeerAddress) }
         }
+      }.thenPeek {
+        discoveryService?.start()
+      }.thenPeek {
+        maruPeerManagerService.start(discoveryService, p2pNetwork)
       }.thenPeek {
```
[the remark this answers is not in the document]
pinges: done :-)

p2p/src/main/kotlin/maru/p2p/P2PNetworkImpl.kt
```diff
@@ -105,6 +131,10 @@ class P2PNetworkImpl(
             .createPeerAddress(peer)
             ?.let { address -> addStaticPeer(address as MultiaddrPeerAddress) }
         }
+      }.thenPeek {
+        discoveryService?.start()
+      }.thenPeek {
+        maruPeerManagerService.start(discoveryService, p2pNetwork)
```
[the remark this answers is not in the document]
pinges: At the moment it will make sure we do only connect to max peers.

p2p/src/test/kotlin/maru/p2p/discovery/MaruDiscoveryServiceTest.kt
```diff
@@ -0,0 +1,130 @@
...
+        },
+    )
+  private val dummyPrivKey = ByteArray(32) { 1 }
+  private val dummyNodeId = Bytes.of(1, 2, 3, 4, 5, 6, 7, 8, 9, 0)
+  private val dummyAddr = Optional.of(InetSocketAddress(InetAddress.getByName("1.1.1.1"), 1234))
+
+  @BeforeEach
+  fun setUp() {
+    val p2pConfig = mock(maru.config.P2P::class.java)
+    `when`(p2pConfig.ipAddress).thenReturn("127.0.0.1")
```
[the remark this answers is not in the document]
pinges: nice!

p2p/src/test/kotlin/maru/p2p/messages/StatusHandlerTest.kt
```diff
@@ -27,21 +28,21 @@ class StatusHandlerTest {
...
   fun `responds with current status`() {
     val beaconChain = mock<BeaconChain>()
-    val forkIdHashProvider = mock<ForkIdHashProvider>()
+    val forkIdBytesProvider = mock<() -> Bytes>()
     val latestBeaconState = DataGenerators.randomBeaconState(0U)
-    val forkIdHash = Random.nextBytes(32)
+    val forkIdHash = Bytes32.random()
     whenever(beaconChain.getLatestBeaconState()).thenReturn(latestBeaconState)
-    whenever(forkIdHashProvider.currentForkIdHash()).thenReturn(forkIdHash)
+    whenever(forkIdBytesProvider.invoke()).thenReturn(forkIdHash)
```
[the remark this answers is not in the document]
pinges: done

p2p/src/main/kotlin/maru/p2p/MaruPeerManagerService.kt
pinges: the exception was thrown during the discovery process. getKnownPeers() returns the already discovered peers. 
Because of the way discv5 works, the searchForNewPeers() does not return the newly found peers. We always have to call getKnownPeers() to get all the discovered peers, including the new ones. 
If searchForNewPeers() is throwing an exception we might already have discovered new peers, so we call getKnownPeers() and try to connect to them.

p2p/src/main/kotlin/maru/p2p/MaruPeerManagerService.kt
pinges: Yeah, I guess the code would look nicer, but we would loose the chance to connect earlier? Not sure ...

p2p/src/main/kotlin/maru/p2p/MaruPeerManagerService.kt
pinges: I thought it should be safe ...
Can you point me to where you are doing something similar in LineaFinalizationProvider? Could not see it :-)
With regards to a delay, I guess it depends on how important it is to find your peers, compared to other stuff that is going on at start-up. Without peers you can't really do anything, but maybe it is OK to start with less than max peers and keep on searching? Or maybe have a min peers? After min peers are found we slow down?

p2p/src/main/kotlin/maru/p2p/MaruPeerManagerService.kt
```diff
@@ -0,0 +1,133 @@
...
+import tech.pegasys.teku.networking.p2p.peer.Peer
+
+class MaruPeerManagerService(
+  p2pConfig: P2P,
+) : PeerHandler {
+  private val log: Logger = LogManager.getLogger(this::class.java)
+
+  private val maxPeers = p2pConfig.maxPeers
+  private val currentlySearching = AtomicBoolean(false)
+  private val connectionInProgress = mutableListOf<Bytes>()
```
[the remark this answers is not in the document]
pinges: I guess with a ByteArray the contains might be a bit faster, but would this have other benefits?

p2p/src/main/kotlin/maru/p2p/MaruPeerManagerService.kt
```diff
@@ -0,0 +1,133 @@
...
+              if (throwable != null) {
+                log.debug("Failed to discover peers: {}\n{}", throwable.message, throwable.stackTraceToString())
+                discoveryService.getKnownPeers().forEach { peer ->
+                  tryToConnectIfNotFull(peer)
+                }
+              } else {
+                availablePeers.forEach { peer ->
+                  tryToConnectIfNotFull(peer)
+                }
+              }
```
Filter94, line 72:
```suggestion
              if (throwable != null) {
                log.debug("Failed to discover peers: {}\n{}", throwable.message, throwable.stackTraceToString())
                discoveryService.getKnownPeers()
              } else {
                availablePeers
              }.forEach { peer ->
                tryToConnectIfNotFull(peer)
              }
```

p2p/src/main/kotlin/maru/p2p/MaruPeerManagerService.kt
Filter94: But why do we reconnect to known peers? Shouldn't we just do nothing and connect to newly discovered peers on the next iteration when `searchForPeers` hopefully doesn't throw an exception?

Filter94: Also, what happens when we try to connect to a peer that we're already connected to?

p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
```diff
@@ -0,0 +1,183 @@
...
+  fun updateForkId(forkId: ForkId) { // TODO: Need to call this when the fork id changes
+    discoverySystem.updateCustomFieldValue(
+      FORK_ID_FIELD_NAME,
+      forkId.bytes,
+    )
+  }
+
+  fun searchForPeers(): SafeFuture<Collection<MaruDiscoveryPeer>> =
+    SafeFuture
+      .of(discoverySystem.searchForNewPeers())
```
Filter94, line 110: You can use `toSafeFuture()` extension from `implementation "build.linea.internal:futures"`

p2p/src/main/kotlin/maru/p2p/MaruPeerManagerService.kt
```diff
@@ -0,0 +1,133 @@
...
+        if (stopCalled) {
+          return
+        }
+        if (p2pNetwork.peerCount >= maxPeers || connectionInProgress.contains(peer.nodeIdBytes)) {
+          return
+        }
+        connectionInProgress.add(peer.nodeIdBytes!!)
+      }
+
+      log.debug("Peer {} Connecting to peer {}...", discoveryService?.getLocalNodeRecord()?.nodeId, peer.nodeIdBytes)
```
Filter94, line 111: If execution got here, it means that `discoveryService` isn't null. So we better make it clear for the reader of the code by using !!

p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
```diff
@@ -0,0 +1,183 @@
...
+
+  fun getKnownPeers(): Collection<MaruDiscoveryPeer> =
+    discoverySystem
+      .streamLiveNodes()
+      .map { node: NodeRecord ->
+        convertNodeRecordToDiscoveryPeer(node)
+      }.filter { checkPeer(it) }
+      .toList()
+
+  fun getLocalNodeRecord(): NodeRecord = discoverySystem.localNodeRecord
```
Filter94, line 122: The way it's used, we're better off making it a val

p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
```diff
@@ -0,0 +1,183 @@
...
+          p2pConfig.discoveryPort.toInt(),
+          p2pConfig.port.toInt(),
+        ).customField(FORK_ID_FIELD_NAME, forkIdProvider.invoke().bytes)
+    // TODO: do we want more custom fields to identify version/topics/role/something else?
+
+    return nodeRecordBuilder.build()
+  }
+
+  private fun checkPeer(peer: MaruDiscoveryPeer): Boolean {
+    if (peer.nodeIdBytes == null ||
```
Filter94, line 164: I don't see a point of this. Can't we avoid this check by making all the fields non-optional?

p2p/src/test/kotlin/maru/p2p/DiscoveryTest.kt
Filter94: So maybe we should invest some time to allocate port dynamically?

p2p/src/test/kotlin/maru/p2p/DiscoveryTest.kt
```diff
@@ -0,0 +1,129 @@
...
+import maru.p2p.discovery.MaruDiscoveryService
+import org.apache.tuweni.bytes.Bytes
+import org.assertj.core.api.Assertions.assertThat
+import org.awaitility.Awaitility
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.parallel.Execution
+import org.junit.jupiter.api.parallel.ExecutionMode
+
+@Execution(ExecutionMode.SAME_THREAD)
+class DiscoveryTest {
```
Filter94, line 23: Also, how this class is different from `MaruDiscoveryServiceTest`? It seems that they both test `MaruDiscoveryService`

p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
```diff
@@ -0,0 +1,183 @@
...
+
+  private var discoverySystem: DiscoverySystem
+
+  private val privateKey = SecretKeyParser.fromLibP2pPrivKey(Bytes.wrap(privateKeyBytes))
+
+  private val bootnodes =
+    p2pConfig.bootnodes
+      .stream()
+      .map { NodeRecordFactory.DEFAULT.fromEnr(it) }
+      .toList()
```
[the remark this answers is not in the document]
pinges: It is a val?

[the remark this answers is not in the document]
pinges: Done

p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
```diff
@@ -0,0 +1,183 @@
...
+
+    discoveryNetworkBuilder.listen(p2pConfig.ipAddress, p2pConfig.discoveryPort.toInt())
+    discoveryNetworkBuilder.secretKey(privateKey)
+    discoveryNetworkBuilder.localNodeRecord(localNodeRecord())
+    discoveryNetworkBuilder.bootnodes(bootnodes)
+
+    discoverySystem = discoveryNetworkBuilder.build()
+  }
+
+  fun start() {
```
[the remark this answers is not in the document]
Filter94: Unless you have something that wasn't checked in, `thenRun` returns a future that probably won't complete by the time `start` returns

p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
pinges: done

p2p/src/main/kotlin/maru/p2p/MaruPeerManagerService.kt
pinges: yes

p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
pinges: The node record that is used to create the MaryDiscoveryPeer is created from data sent to us from the other peer.  They can send anything.

Filter94: Maybe instead of a map after `streamLiveNodes` we can use flatMap and reject all the items that don't make sense for us

p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
```diff
@@ -0,0 +1,183 @@
...
+      .map { NodeRecordFactory.DEFAULT.fromEnr(it) }
+      .toList()
+
+  val delayedExecutor: AsyncRunner =
+    ScheduledExecutorAsyncRunner.create(
+      "DiscoveryService",
+      1,
+      1,
+      5,
+      MetricTrackingExecutorFactory(NoOpMetricsSystem()),
```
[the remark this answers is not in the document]
pinges: Passing down a NoOpMetricsSystem from the P2PBetworkImpl to both, the factory and the discovery system.

p2p/src/main/kotlin/maru/p2p/MaruPeerManagerService.kt
pinges: nice

p2p/src/main/kotlin/maru/p2p/MaruPeerManagerService.kt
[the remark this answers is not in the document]
pinges: So you mean that I should convert the Bytes that I get from the NodeRecord to a ByteArray, so I can convert it back to Bytes when I need them for the DiscoveryPeer?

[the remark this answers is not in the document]
Filter94: 1. We could turn all the `Bytes` fields in `MaruDiscoveryPeer` into `ByteArray`-s. Then this list will be an `mutableListOf<ByteArray>()`. But since `MaruDiscoveryPeer` is a wrapper around `DiscoveryPeer`, I get why you would might want to stick to `Bytes`. But in this regard, do we really want `MaruDiscoveryPeer` wrapper? What value does it bring? And why won't we put fork id into the `enrForkId` that is in `DiscoveryPeer`? And now it begs other question. Why are we using `DiscoveryPeer` at all?

2. Maintaining a list of `MaruDiscoveryPeer` might make more sense. It would fix an edge case bug when there are 2 peers with the same id, but with different `ForkId`-s.

config/src/main/kotlin/maru/config/Config.kt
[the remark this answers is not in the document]
pinges: I really like that :-)

jframe (approved):
I think this is fine to go in. We can create issues to fix any outstanding comments

p2p/src/main/kotlin/maru/p2p/MaruPeerManager.kt
```diff
@@ -50,9 +121,47 @@ class MaruPeerManager(
 
   override fun onDisconnect(peer: Peer) {
     connectedPeers.remove(peer.id)
+    if (!stopCalled && p2pNetwork.peerCount < maxPeers) {
```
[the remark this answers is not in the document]
pinges: I prefer to do that this way ...

p2p/src/main/kotlin/maru/p2p/MaruPeerManager.kt
```diff
@@ -23,11 +36,69 @@ private const val STATUS_TIMEOUT_SECONDS = 10L
...
 ) : PeerHandler,
   PeerLookup {
+  private val log: Logger = LogManager.getLogger(this::class.java)
+  private val maxPeers = p2pConfig.maxPeers
+  private val currentlySearching = AtomicBoolean(false)
+  private val connectionInProgress = mutableListOf<Bytes>()
+
+  private var discoveryService: MaruDiscoveryService? = null
+  private lateinit var p2pNetwork: tech.pegasys.teku.networking.p2p.network.P2PNetwork<Peer>
+  private var stopCalled = false
```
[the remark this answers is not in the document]
pinges: changed it to volatile, as it could be read by multiple threads, which should be safe, as it is only ever changed to true, never to false.

p2p/src/main/kotlin/maru/p2p/MaruPeerManager.kt
```diff
@@ -23,11 +36,69 @@ private const val STATUS_TIMEOUT_SECONDS = 10L
...
+    stopCalled = true
+    return SafeFuture.completedFuture<Unit>(null)
+  }
+
+  private fun searchForPeersUntilMaxReached() {
+    if (!stopCalled && currentlySearching.compareAndSet(false, true)) {
+      discoveryService?.let { discoveryService ->
+        discoveryService
+          .searchForPeers()
+          .orTimeout(Duration.ofSeconds(30L))
```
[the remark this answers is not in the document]
pinges: I'm not a big fan to make everything configurable, but if you think we might want to change that timeout, we should do that.

app/src/main/kotlin/maru/app/MaruAppFactory.kt
```diff
@@ -86,24 +83,15 @@ class MaruAppFactory {
...
-    val forkIdHashProvider =
-      ForkIdHashProvider(
-        chainId = beaconGenesisConfig.chainId,
-        beaconChain = beaconChain,
-        forksSchedule = beaconGenesisConfig,
-        forkIdHasher = forkIdHasher,
+    val forkId =
+      ForkId(
+        chainId = 1L.toUInt(),
+        genesisRootHash = ByteArray(32),
```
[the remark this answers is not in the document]
pinges: done

p2p/src/test/kotlin/maru/p2p/discovery/MaruDiscoveryServiceTest.kt
```diff
@@ -0,0 +1,279 @@
...
+  private val dummyNodeId = Bytes.of(1, 2, 3, 4, 5, 6, 7, 8, 9, 0)
+  private val dummyAddr = Optional.of(InetSocketAddress(InetAddress.getByName("1.1.1.1"), 1234))
+
+  @BeforeEach
+  fun setUp() {
+    val p2pConfig = mock(P2P::class.java)
+    whenever(p2pConfig.ipAddress).thenReturn("127.0.0.1")
+    whenever(p2pConfig.discoveryPort).thenReturn(9000.toUInt())
+    whenever(p2pConfig.port).thenReturn(9001.toUInt())
+    whenever(p2pConfig.bootnodes).thenReturn(listOf())
```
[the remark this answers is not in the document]
pinges: removed mocking

app/src/main/kotlin/maru/app/MaruAppFactory.kt
```diff
@@ -219,6 +222,8 @@ class MaruAppFactory {
           nextExpectedBeaconBlockNumber = nextExpectedBeaconBlockNumber,
           metricsFacade = metricsFacade,
           statusMessageFactory = statusMessageFactory,
+          metricsSystem = NoOpMetricsSystem(),
```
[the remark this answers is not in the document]
pinges: fixed

p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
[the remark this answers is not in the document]
pinges: I have added a get()

p2p/src/main/kotlin/maru/p2p/MaruPeerManager.kt
```diff
@@ -23,11 +36,69 @@ private const val STATUS_TIMEOUT_SECONDS = 10L
...
+  ) {
+    this.discoveryService = discoveryService
+    this.p2pNetwork = p2pNetwork
+    searchForPeersUntilMaxReached()
+  }
+
+  fun stop(): SafeFuture<Unit> {
+    stopCalled = true
+    return SafeFuture.completedFuture<Unit>(null)
+  }
```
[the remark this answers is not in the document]
pinges: it is not longer ...

Issue #97, opened 2025-05-09
Observability metrics


gauravahuja:
Publish maru metrics to be used in grafana charts and for alerting purposes. These metrics are related to 
1. Beacon Node
  - version
  - block height
  - p2p peers
  - round latency
  - etc
2. System/JVM metrics

PR #134, opened 2025-06-03
Add observability options and setup metrics registry

gauravahuja:
#97 
Depends on https://github.com/Consensys/linea-monorepo/pull/1097

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,5 @@ out/
 
 # Generated files
 docker/initialization/*.json
+
+.java-version
--- a/app/src/main/kotlin/maru/app/MaruApp.kt
+++ b/app/src/main/kotlin/maru/app/MaruApp.kt
@@ -31,17 +31,14 @@ import maru.consensus.Web3jMetadataProvider
 import maru.consensus.blockimport.FollowerBeaconBlockImporter
 import maru.consensus.blockimport.NewSealedBeaconBlockHandlerMultiplexer
 import maru.consensus.delegated.ElDelegatedConsensusFactory
-import maru.consensus.state.FinalizationState
-import maru.core.BeaconBlockBody
+import maru.consensus.state.FinalizationProvider
+import maru.consensus.state.InstantFinalizationProvider
 import maru.core.Protocol
 import maru.crypto.Crypto
 import maru.database.kv.KvDatabaseFactory
-import maru.p2p.NoOpP2PNetwork
 import maru.p2p.P2PNetwork
-import maru.p2p.P2PNetworkImpl
 import maru.p2p.SealedBeaconBlockBroadcaster
 import maru.p2p.ValidationResult
-import maru.serialization.rlp.RLPSerializers
 import net.consensys.linea.async.get
 import net.consensys.linea.metrics.Tag
 import net.consensys.linea.metrics.micrometer.MicrometerMetricsFacade
@@ -52,14 +49,15 @@ import org.apache.logging.log4j.Logger
 import org.hyperledger.besu.metrics.noop.NoOpMetricsSystem
 import org.hyperledger.besu.plugin.services.metrics.MetricCategory
 import tech.pegasys.teku.infrastructure.async.SafeFuture
-import tech.pegasys.teku.networking.p2p.network.config.GeneratingFilePrivateKeySource
 
 class MaruApp(
   val config: MaruConfig,
   beaconGenesisConfig: ForksSchedule,
   clock: Clock = Clock.systemUTC(),
   // This will only be used if config.p2pConfig is undefined
-  private var p2pNetwork: P2PNetwork = NoOpP2PNetwork,
+  private var p2pNetwork: P2PNetwork,
+  private val privateKeyProvider: () -> ByteArray,
+  private val finalizationProvider: FinalizationProvider = InstantFinalizationProvider,
 ) : AutoCloseable {
   private val log: Logger = LogManager.getLogger(this::javaClass)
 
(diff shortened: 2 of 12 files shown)


executionlayer/src/main/kotlin/maru/executionlayer/client/PragueWeb3JJsonRpcExecutionLayerEngineApiClient.kt
```diff
@@ -32,28 +35,66 @@ import tech.pegasys.teku.infrastructure.bytes.Bytes8
...
-      when {
-        it.payload != null ->
-          Response.fromPayloadReceivedAsJson(it.payload.executionPayload.toDomainExecutionPayload())
+    timerProvider
+      .withTags(listOf(Tag("method", "getPayload")))
+      .captureTime(web3jEngineClient.getPayloadV4(payloadId))
+      .thenApply {
+        when {
+          it.payload != null -> {
+            counterProvider
```
Filter94, line 48: You don't really need counters as a separate thing. Counts are captured by timers themselves

buildSrc/src/main/groovy/maru.kotlin-common-minimal-conventions.gradle
```diff
@@ -11,6 +11,7 @@ java {
 }
 
 repositories {
+  mavenLocal()
```
Filter94, line 14: I feel like it's safer to comment it out and enable it only for the local debugging. I don't think it should go into the repo

buildSrc/src/main/groovy/maru.kotlin-common-minimal-conventions.gradle
```diff
@@ -49,6 +50,7 @@ dependencies {
   // check later on if this may raise incompatibilities
   testImplementation 'org.mockito.kotlin:mockito-kotlin'
   testImplementation 'org.awaitility:awaitility-kotlin'
+  testImplementation 'build.linea.internal:micrometer'
```
Filter94, line 53: I don't think micrometer metrics should go here

buildSrc/src/main/groovy/maru.kotlin-common-conventions.gradle
```diff
@@ -14,6 +14,7 @@ dependencies {
   } because("We are using Teku's SafeFuture, an improved/extension of CompletableFuture.")
 
   implementation "com.michael-bull.kotlin-result:kotlin-result"
+  implementation "build.linea.internal:metrics"
```
Filter94, line 17: I wouldn't put it here, because I think that it won't be that many gradle modules that will expose any metrics

jvm-libs/extensions/build.gradle
```diff
@@ -16,3 +16,8 @@
 plugins {
   id 'maru.kotlin-library-conventions'
 }
+
+dependencies {
+  api "build.linea.internal:futures"
+  api "tech.pegasys.teku.internal:executionclient"
```
Filter94, line 22: TBH, exposing these as `api` doesn't seem like a good idea, as it will blur the dependencies. I would rather avoid transitive dependencies on `build.linea.internal:futures` and import them explicitly as `implementation` where they're needed

jvm-libs/extensions/src/main/kotlin/TimerExtensions.kt
```diff
@@ -0,0 +1,20 @@
...
+ * SPDX-License-Identifier: MIT OR Apache-2.0
+ */
+package maru.extensions
+
+import net.consensys.linea.async.toSafeFuture
+import net.consensys.linea.metrics.DynamicTagTimer
+import net.consensys.linea.metrics.Timer
+import tech.pegasys.teku.infrastructure.async.SafeFuture
+
+fun <T> Timer.captureTimeSafeFuture(safeFuture: SafeFuture<T>): SafeFuture<T> =
```
Filter94, line 16: Since `CompletableFuture` is easily convertable to `SafeFuture`, why won't we convert the argument where it's used instead? Otherwise, by the same reasoning, instead of converting the `CompletableFuture`-s to `SafeFuture` we would add extensions for other usecases of `SafeFuture`, which doesn't scale well and will bloat the API.

executionlayer/src/main/kotlin/maru/executionlayer/client/PragueWeb3JJsonRpcExecutionLayerEngineApiClient.kt
```diff
@@ -10,22 +10,51 @@ package maru.executionlayer.client
...
 import tech.pegasys.teku.infrastructure.bytes.Bytes8
 
 class PragueWeb3JJsonRpcExecutionLayerEngineApiClient(
-  private val web3jEngineClient: Web3JExecutionEngineClient,
+  private val web3jClient: Web3JClient,
+  private val metricsFacade: MetricsFacade,
 ) : ExecutionLayerEngineApiClient {
+  private val web3jEngineClient: Web3JExecutionEngineClient = Web3JExecutionEngineClient(web3jClient)
+
+  private fun <T> createRequestTimer(method: String): DynamicTagTimer<Response<T>> =
```
Filter94, line 38: It's not used ATM, so I guess this  part needs to be completed

jvm-libs/extensions/src/main/kotlin/TimerExtensions.kt
Filter94: I would suggest to either opt for `timer.captureTime(sf.toCompletableFuture()).toSafeFuture()` or to rename the extension you added to 
```
Timer.captureTime(safeFuture: SafeFuture<T>): SafeFuture<T>
```
I don't see a reason to add the type info into the method name itself, when the type info is in the argument and the return type.

buildSrc/src/main/groovy/maru.kotlin-common-minimal-conventions.gradle
gauravahuja: Will make those changes in a separate PR. I tried to do those and was getting CI issues with groove spotless. Did not have those issues locally.

jvm-libs/extensions/src/main/kotlin/TimerExtensions.kt
Filter94: Ah, ok. Now it makes sense

Filter94 (approved):

Issue #135, opened 2025-06-03
Fork Id


Filter94:
## Description
We need to add a way to compute fork ID. Fork id has to be the same for nodes that are participating on the same network and on the same fork and it shouldn't match for the nodes that aren't participating in the network

## Motivation
This will help on peering with the right peers

## Tasks
- [ ] Carve out the configurations out of Maru's genesis file that will be static for the next fork 
- [ ] Implement Fork ID calculation logic. It should include the chainId, the contents of the Maru's genesis file for the current fork and Beacon Chain genesis block hash

## Acceptance criteria
- [ ] Fork Id calculation logic is implemented and it's covered with tests

## Risks
- [ ] If we define the fork Id incorrectly, the nodes might not find right peers and if it's too lose, there's a risk to peer with the nodes we're not interested in

## Remember to
 - [ ] Add the `documentation` label in case there is an impact on the documentation
 - [ ] Add `priority` and `team` labels
 - [ ] Add Task for updating the Runbook or adding/updating existing metrics and alerts.

PR #137, opened 2025-06-04
[135] Added logic for fork id computation

Filter94:
Contains work on #135

* Moved consensus configs to config module along with Beacon chain config parsing logic
* Renamed `Serializer` interface to `SerDe`, introduced `Serializer` and `Deserializer` as separate interfaces
* Moved `ConsensusConfig` into core
* Added `ForkId` and `ForkIdHasher` to core
* Added `QbftForkIdComputer` implementation to p2p module
* Added serializers to serialize `ForkId` and `QbftConsensusConfig`
* Removed `KeccakHasher` in favour of a function reference


serialization/src/main/kotlin/maru/serialization/ForkIdSerializers.kt
```diff
@@ -0,0 +1,52 @@
...
+
+  object QbftForkIdSerializer : Serializer<ForkId<QbftConsensusConfig>> {
+    override fun serialize(value: ForkId<QbftConsensusConfig>): ByteArray {
+      val serializedConsensusConfig = QbftConsensusConfigSerializer.serialize(value.consensusConfig)
+
+      val buffer =
+        ByteBuffer
+          .allocate(4 + serializedConsensusConfig.size + 32)
+          .putInt(value.chainId.toInt())
+          .put(serializedConsensusConfig)
```
jframe, line 46: don't think we need the config in addition to the `genesisRootHash` as this also has the validators.

core/src/main/kotlin/maru/consensus/ForkId.kt
```diff
@@ -0,0 +1,52 @@
...
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+ */
+package maru.consensus
+
+import maru.core.Hasher
+import maru.serialization.Serializer
+
+data class ForkId<T : ConsensusConfig>(
```
jframe, line 21: Not sure why we need this data type and the serializer. I was thinking the forkId calculation would be quite simple with something like `forkId = Hasher.hash(elFork.ordinal ++ chainId ++ genesisRootHash)`.

Filter94: I think your approach is not going to be future proof because it doesn't include the data from the future fork configurations. We don't know what that data is going to be, so by introducing Fork ID with a `ConsensusConfig` as a field. If we're missing a ForkId computation for the future fork with say `QbftConsensusConfigV2`, it will fail in compile time. So by introducing this type we're introducing some additional type safety for the future extensions

config/src/main/kotlin/maru/config/consensus/qbft/QbftConsensusConfig.kt
```diff
@@ -6,13 +6,18 @@
...
 
 data class QbftConsensusConfig(
   val validatorSet: Set<Validator>,
   val elFork: ElFork,
-) : ConsensusConfig
+) : ConsensusConfig {
+  companion object {
+    enum class ElFork {
+      Prague,
+    }
```
gauravahuja, line 21: I feel like the ELfork enum should be global and not defined inside QbftConsensusConfig

serialization/src/test/kotlin/maru/serialization/ForkSpecSerializerTest.kt
```diff
@@ -0,0 +1,121 @@
...
+import maru.config.consensus.qbft.QbftConsensusConfig
+import maru.config.consensus.qbft.QbftConsensusConfig.Companion.ElFork
+import maru.consensus.ForkSpec
+import maru.core.ext.DataGenerators
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+
+class ForkSpecSerializerTest {
+  @Test
+  fun `serialization is deterministic for same input`() {
```
gauravahuja, line 20: Can you add a test for ElDelegatedConfig?

gauravahuja (commented):
Left a few comments. Generally it would have been simpler to move the unrelated Serializer rename to SerDe in a different PR.

serialization/src/test/kotlin/maru/serialization/ForkSpecSerializerTest.kt
Filter94: No, because it's not supposed to be serialized for the hashing

p2p/src/main/kotlin/maru/p2p/QbftForkIdComputer.kt
```diff
@@ -6,11 +6,15 @@
...
 
-import maru.core.Hasher
+import maru.consensus.ForkIdHasher
 import maru.crypto.Hashing
+import maru.serialization.ForkIdSerializers
 
-object KeccakHasher : Hasher {
-  override fun hash(serializedBytes: ByteArray): ByteArray = Hashing.keccak(serializedBytes)
-}
+val QbftForkIdComputer: ForkIdHasher =
```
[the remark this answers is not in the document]
Filter94: Yes, that's true. I'll rename it

gauravahuja (approved):

PR #143, opened 2025-06-12
move FinalizationProvider typings into core

fluentcrafter:
This PR does 2 key changes:

- moves FinalizationState into core and creates FinalizationProvider interface
- Creates MaruAppFactory to facilitate MaruApp configuration. The goal is for the log run for MaruApp to have only interfaces in its constructor and not config. This facilitates testing

### DIFF
--- a/app/src/test/kotlin/maru/app/MaruFollowerNegativeTest.kt
+++ b/app/src/test/kotlin/maru/app/MaruFollowerNegativeTest.kt
@@ -53,7 +53,7 @@ class MaruFollowerNegativeTest {
           ethereumJsonRpcUrl = ethereumJsonRpcBaseUrl,
           engineApiRpc = engineRpcUrl,
           dataDir = tmpDir,
-          p2pNetwork = spyingP2PNetwork,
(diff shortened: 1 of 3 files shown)


Filter94 (approved):

app/src/main/kotlin/maru/app/MaruApp.kt
```diff
@@ -27,67 +27,38 @@ import maru.consensus.Web3jMetadataProvider
...
 import tech.pegasys.teku.infrastructure.async.SafeFuture
-import tech.pegasys.teku.networking.p2p.network.config.GeneratingFilePrivateKeySource
 
 class MaruApp(
   config: MaruConfig,
   beaconGenesisConfig: ForksSchedule,
   clock: Clock = Clock.systemUTC(),
   // This will only be used if config.p2pConfig is undefined
-  private var p2pNetwork: P2PNetwork = NoOpP2PNetwork,
+  private var p2pNetwork: P2PNetwork,
```
gauravahuja, line 48: This can be val. I don't see it being reassigned anywhere.

PR #147, opened 2025-06-17
Status message type

### LINKED ISSUE #133: Status message
Implement a status message that will be used when a node first connects. This status message has enough details so a node knows whether it is on the same fork and what it's current head state and block number are.

jframe:
Implements a status message that will be used when a node first connects.

fixes #133

- Modify RPC handling so that specific RPC handlers can be provided
- Split the `MessageType` into two types `GossipMessageType` and `RpcMessageType` with a common type so that gossip handling doesn't need to handle messages that are sent only by rpc and vice versa
- Created a `ForkIdHashProvider` so the current forkId hash can be returned for use in the status message
- Created a status message and a handler that returns the current status


p2p/src/main/kotlin/maru/p2p/P2PNetworkImpl.kt
```diff
@@ -100,24 +104,39 @@ class P2PNetworkImpl(
...
+  ): SafeFuture<*> {
+    val rpcMethodRecord =
+      rpcMethods[message.type.toEnum()] ?: throw IllegalArgumentException("Unsupported message type: ${message.type}")
+    val peerId = peer.id.toString()
+    val request: Bytes = Bytes.wrap(rpcMethodRecord.serDe.serialize(message))
+    val responseHandler = MaruRpcResponseHandler()
+    return sendRequest(peerId, rpcMethodRecord.rpcMethod, request, responseHandler)
+      .thenCompose {
+        responseHandler.response()
+      }.thenApply { bytes -> rpcMethodRecord.serDe.deserialize(bytes.toArrayUnsafe()) }
```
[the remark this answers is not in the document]
jframe: done

p2p/src/main/kotlin/maru/p2p/P2PNetworkImpl.kt
```diff
@@ -109,15 +113,33 @@ class P2PNetworkImpl(
...
+    message: TRequest,
+    peer: Peer,
+  ): SafeFuture<TResponse> {
+    val rpcMethod =
+      rpcMethods[message.type] ?: throw IllegalArgumentException("Unsupported message type: ${message.type}")
+    val peerId = peer.id.toString()
+    val responseHandler = MaruRpcResponseHandler<TResponse>()
+    return sendRequest(
+      peerId,
+      rpcMethod as MaruRpcMethod<TRequest, TResponse>,
```
fluentcrafter, line 135: Why do we need this cast? 

Can we make `private val rpcMethods: MaruRpcMethod<TRequest, TResponse>` typed instead?

fluentcrafter (commented):
LGT, just left some minor suggestions

p2p/src/main/kotlin/maru/p2p/P2PNetworkImpl.kt
jframe: I couldn't it to work without the cast. Would you be ok with waiting for this to be fixed in another PR? I've done some refactoring around this area in my next PR and have been able to remove the cast.

Issue #95, opened 2025-05-09
Node identity APIs


gauravahuja:
These are endpoints to query node related information
https://github.com/ethereum/beacon-APIs/tree/master/apis/node

1. /eth/v1/node/identity
2. /eth/v1/node/peers
3. /eth/v1/node/peers/{peer_id}
4. /eth/v1/node/peer_count
5. /eth/v1/node/version
6. /eth/v1/node/syncing
7. /eth/v1/node/health

PR #158, opened 2025-06-24
Create ApiServer

gauravahuja:
#95 

In this PR
- Create ApiServer that starts a JsonRpc Http server
- Implement NodeGetNetworkIdentity method.


api/src/test/kotlin/maru/api/ApiServerTest.kt
```diff
@@ -0,0 +1,107 @@
...
+        p2pAddresses = NoOpP2PNetwork.nodeAddresses,
+        discoveryAddresses = NoOpP2PNetwork.discoveryAddresses,
+        metadata =
+          Metadata(
+            seqNumber = "1",
+            attnets = emptyList(),
+            syncnets = emptyList(),
+          ),
+      )
+  }
```
fluentcrafter, line 106: this test seems incomplete to me, expectedResponse is not used and we dont' validate the response

fluentcrafter (approved):

api/src/main/kotlin/maru/api/NodeGetNetworkIdentityV1.kt
```diff
@@ -0,0 +1,73 @@
...
+  val enr: String,
+  @JsonProperty("p2p_addresses") val p2pAddresses: List<String>,
+  @JsonProperty("discovery_addresses") val discoveryAddresses: List<String>,
+  val metadata: Metadata,
+)
+
+data class Metadata(
+  @JsonProperty("seq_number") val seqNumber: String,
+  val attnets: List<String>,
+  val syncnets: List<String>,
```
jframe, line 43: These will always be empty for Maru. Is this just here for compatability?

api/src/main/kotlin/maru/api/ApiServer.kt
```diff
@@ -0,0 +1,76 @@
...
+  p2PNetwork: P2PNetwork,
+) {
+  data class Config(
+    val port: UInt,
+    val path: String,
+  )
+
+  val methodHandlers: Map<String, JsonRpcRequestHandler> =
+    mapOf(
+      ApiMethod.NODE_GET_NETWORK_IDENTITY_V1.method to NodeGetNetworkIdentityV1(p2PNetwork),
```
jframe, line 37: What route will be used for this request? It doesn't look like it will be `/eth/v1/node/identity` as defined in the spec for getNetworkIdentity

api/src/main/kotlin/maru/api/NodeGetNetworkIdentityV1.kt
```diff
@@ -0,0 +1,73 @@
...
+    jsonRpcRequest: JsonRpcRequest,
+    jsonObject: JsonObject,
+  ): Future<Result<JsonRpcSuccessResponse, JsonRpcErrorResponse>> =
+    Future.succeededFuture(
+      Ok(
+        JsonRpcSuccessResponse(
+          jsonRpcRequest.id,
+          GetNetworkIdentityJsonResponse(
+            peerId = p2PNetwork.nodeId,
+            enr = p2PNetwork.enr ?: "",
```
jframe, line 60: In the case of no enr I think we should be returning null or no enr field. At least that's what Teku does

api/src/main/kotlin/maru/api/NodeGetNetworkIdentityV1.kt
gauravahuja: yes

api/src/main/kotlin/maru/api/ApiServer.kt
gauravahuja: refactored this to be a rest endpoint with route /eth/v1/node/identity

api/src/main/kotlin/maru/api/NodeGetNetworkIdentityV1.kt
gauravahuja: I don't know, the beacon api spec just says a string https://ethereum.github.io/beacon-APIs/#/Node/getNetworkIdentity

api/src/main/kotlin/maru/api/ApiServer.kt
```diff
@@ -0,0 +1,43 @@
...
+        Javalin
+          .create()
+          .get(NodeGetNetworkIdentity.ROUTE, NodeGetNetworkIdentity(networkDataProvider))
+          .start(config.port.toInt())
+    }
+  }
+
+  fun stop() {
+    app?.stop()
+    app = null
```
fluentcrafter, line 39: I would strongly recommend https://www.http4k.org/ for our APIs instead... besised supporting vertex as a backend, it's designed to be more testable with in Memory fakes. In the long run it just pays off.

Since we are just starting it's very little effort the switch ...

gauravahuja: I prefer going with Javalin, javalin has 7.9k stars vs http4k 2.7k stars. Javalin is also used by Teku so there is knowledge of using it within the company.

fluentcrafter: I would not go for the stars count, but rather look into what each of them offers and how it aligns with our goals.
Http4k is newer, hence fewer stars. But was I said, much better tool when it comes to testing and productivity..

Google AI summary on "Javalin vs Http4K"

> In summary: 
> If you're looking for a simple, lightweight, and easy-to-use framework, Javalin might be a good choice. 
> If you prefer a functional approach, value testability, and need a highly flexible toolkit, http4k could be a better fit. 

Chook wisely ;)

gauravahuja: Javalin also makes it easy in terms of OpenApi integration, and we will need that because will have to document the custom fields for maru endpoints.

fluentcrafter: So it does Http4K

> http4k provides [over 100 pre-built integrations](https://www.http4k.org/ecosystem/) for developers to choose from including Observability, JSON, and OpenAPI support.

It's on their documentation: 

- https://www.http4k.org/
- https://www.http4k.org/ecosystem/http4k/reference/contracts/

fluentcrafter (approved):

api/src/main/kotlin/maru/api/NodeGetNetworkIdentity.kt
```diff
@@ -0,0 +1,69 @@
...
+  @JsonProperty("p2p_addresses") val p2pAddresses: List<String>,
+  @JsonProperty("discovery_addresses") val discoveryAddresses: List<String>,
+  val metadata: Metadata,
+)
+
+/**
+ * https://github.com/ethereum/consensus-specs/blob/v1.3.0/specs/phase0/p2p-interface.md#metadata
+ */
+data class Metadata(
+  @JsonProperty("seq_number") val seqNumber: String,
```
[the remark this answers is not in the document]
gauravahuja: Metadata response in the API is a string (an base 10 integer encoded in the string) If it were hex it would be explicitly mentioned in the api [documentation](https://ethereum.github.io/beacon-APIs/#/Node/getNetworkIdentity)

image

api/src/main/kotlin/maru/api/NodeGetNetworkIdentity.kt
```diff
@@ -0,0 +1,69 @@
...
+)
+
+class NodeGetNetworkIdentity(
+  val networkDataProvider: NetworkDataProvider,
+) : Handler {
+  override fun handle(ctx: Context) {
+    val networkIdentity =
+      NetworkIdentity(
+        peerId = networkDataProvider.getNodeId(),
+        enr = networkDataProvider.getEnr() ?: "",
```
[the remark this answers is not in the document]
gauravahuja: Do you mean a string with "null" or "undefined"

enr in the API response is a string, does not say nullable. [documentation](https://ethereum.github.io/beacon-APIs/#/Node/getNetworkIdentity)

image

PR #165, opened 2025-07-03
Implement node apis

gauravahuja:
#95 

Implement the following node apis:

- [/eth/v1/node/peers](https://ethereum.github.io/beacon-APIs/#/Node/getPeers)
- [/eth/v1/node/peers/{peer_id}](https://ethereum.github.io/beacon-APIs/#/Node/getPeer)
- [/eth/v1/node/peer_count](https://ethereum.github.io/beacon-APIs/#/Node/getPeerCount)
- [/eth/v1/node/version](https://ethereum.github.io/beacon-APIs/#/Node/getNodeVersion)
- [/eth/v1/node/syncing](https://ethereum.github.io/beacon-APIs/#/Node/getSyncingStatus)
- [/eth/v1/node/health](https://ethereum.github.io/beacon-APIs/#/Node/getHealth)

Syncing and Health apis have placeholder data for now as they need information about syncing which isn't implemented yet.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -151,13 +151,13 @@ allprojects {
       events 'passed', 'skipped', 'failed'
       showStandardStreams = true
     }
+  }
 
-    systemProperties["junit.jupiter.execution.parallel.enabled"] =  true
+  tasks.withType(Test).configureEach {
+    systemProperties["junit.jupiter.execution.parallel.enabled"] = true
     systemProperties["junit.jupiter.execution.parallel.mode.default"] = "concurrent"
     systemProperties["junit.jupiter.execution.parallel.mode.classes.default"] = "concurrent"
-    systemProperties["junit.jupiter.execution.parallel.config.strategy"] = "fixed"
-    systemProperties["junit.jupiter.execution.parallel.config.fixed.parallelism"] =
-      Runtime.runtime.availableProcessors().intdiv(2) ?: 1
+    maxParallelForks = 1
   }
 }
 
@@ -190,7 +190,6 @@ subprojects {
   }
 }
 
-
 task jacocoAggregatedReport(type: JacocoReport) {
   group = "verification"
 
@@ -204,7 +203,6 @@ task jacocoAggregatedReport(type: JacocoReport) {
     html.destination file("build/reports/jacocoHtml")
   }
 }
-
 
 apply plugin: 'application'
 
--- a/build.gradle
+++ b/build.gradle
@@ -151,13 +151,13 @@ allprojects {
       events 'passed', 'skipped', 'failed'
       showStandardStreams = true
     }
-  }
 
-  tasks.withType(Test).configureEach {
-    systemProperties["junit.jupiter.execution.parallel.enabled"] = true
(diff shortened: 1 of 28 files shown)


build.gradle
```diff
@@ -151,13 +151,13 @@ allprojects {
       events 'passed', 'skipped', 'failed'
       showStandardStreams = true
     }
-  }
 
-  tasks.withType(Test).configureEach {
-    systemProperties["junit.jupiter.execution.parallel.enabled"] = true
+    systemProperties["junit.jupiter.execution.parallel.enabled"] =  true
```
Filter94, line 155: Let's address it in a separate PR. E.g. [here](https://github.com/Consensys/maru/pull/184)

gradle.properties
```diff
@@ -1,12 +1,9 @@
-org.gradle.parallel=true
-org.gradle.caching=true
+# org.gradle.parallel=true
+# org.gradle.caching=true
```
Filter94, line 2: Same here, let's address it in a different PR

p2p/src/main/kotlin/maru/p2p/P2PNetworkImpl.kt
```diff
@@ -31,8 +31,8 @@ import tech.pegasys.teku.networking.p2p.libp2p.PeerAlreadyConnectedException
 import tech.pegasys.teku.networking.p2p.network.PeerAddress
 import tech.pegasys.teku.networking.p2p.peer.DisconnectReason
 import tech.pegasys.teku.networking.p2p.peer.NodeId
-import tech.pegasys.teku.networking.p2p.peer.Peer
 import tech.pegasys.teku.networking.p2p.rpc.RpcStreamController
+import tech.pegasys.teku.networking.p2p.peer.Peer as TekuPeer
 
 class P2PNetworkImpl(
```
Filter94, line 37: This probably needs some coordination with Stefan's [PR](https://github.com/Consensys/maru/pull/126)

Filter94 (commented):
Looks good overall. Once the syncing endpoint is properly implemented it would be good to improve the coverage for it, because it's one of the most important methods here

Filter94 (dismissed):

core/src/main/kotlin/maru/p2p/P2PNetwork.kt
```diff
@@ -99,4 +99,28 @@ interface P2PNetwork {
   val discoveryAddresses: List<String>
 
   val enr: String?
+
+  fun getPeers(): List<PeerInfo>
+
+  fun getPeer(peerId: String): PeerInfo?
```
jframe, line 105: These are very similar to the methods on `PeerLookup` except using peerId instead of nodeId. A better solution I think would be for the `P2PNetworkDataProvider` to use the `PeerLookup` and do the translation and conversion there.

gauravahuja: The issue with that is to instantiate P2PNetworkDataProvider I would need to inject a P2PLookup object. P2P lookup object is privately instantiated inside P2PNetworkImpl. So this approach does not work right away without some refactoring.

jframe (approved):

PR #175, opened 2025-07-08
Send status message when a peer connects

### LINKED ISSUE #152: Send status message when peer is connected

jframe:
This sends a status message when a peer connects so that we know it is valid peer. Note that the validation isn't implemented in this PR for the forkID as this is already implemented in the discovery PR .

fixes #152

* Send a status message when a peer connects
* Disconnect a peer if the peer message isn't sent
* Move the rpc methods onto a `MaruPeer` to make using them simpler as part of syncing


p2p/src/main/kotlin/maru/p2p/MaruPeer.kt
```diff
@@ -0,0 +1,114 @@
...
+import tech.pegasys.teku.networking.p2p.reputation.ReputationAdjustment
+import tech.pegasys.teku.networking.p2p.rpc.RpcMethod
+import tech.pegasys.teku.networking.p2p.rpc.RpcRequestHandler
+import tech.pegasys.teku.networking.p2p.rpc.RpcResponseHandler
+import tech.pegasys.teku.networking.p2p.rpc.RpcStreamController
+
+interface MaruPeer : Peer {
+  fun getStatus(): Status?
+
+  fun sendStatus(): SafeFuture<Status>
```
Filter94, line 29: Should we update status on every block? Or when remote peer's block height is close to Head, we optimistically assume it's actually synced?

Filter94: Maybe in addition to block heights we can add a field whether peer thinks it's synced

Filter94 (dismissed):

p2p/src/main/kotlin/maru/p2p/MaruPeer.kt
jframe: As discussed on the sync call, it's not clear whether we will need this or not. If we do we can this in later

Filter94 (approved):

Issue #168, opened 2025-07-03
Flacky Test - MaruFollowerTest


fluentcrafter:
Job Run - https://github.com/Consensys/maru/actions/runs/16050242388/job/45290938991
```
MaruFollowerTest > Maru follower is able to import blocks after going down() FAILED
    org.opentest4j.AssertionFailedError: 
    expected: 
      [org.web3j.protocol.core.methods.response.EthBlock$Block@3b0a4301,
          org.web3j.protocol.core.methods.response.EthBlock$Block@c8a60cff,
          org.web3j.protocol.core.methods.response.EthBlock$Block@77ccbd9d,
          org.web3j.protocol.core.methods.response.EthBlock$Block@1969d6a4,
          org.web3j.protocol.core.methods.response.EthBlock$Block@6ac0f99c,
          org.web3j.protocol.core.methods.response.EthBlock$Block@c88aeb8d,
          org.web3j.protocol.core.methods.response.EthBlock$Block@d88051c9,
          org.web3j.protocol.core.methods.response.EthBlock$Block@cffff844,
          org.web3j.protocol.core.methods.response.EthBlock$Block@9a0769f,
          org.web3j.protocol.core.methods.response.EthBlock$Block@1e351a0b]
     but was: 
      [org.web3j.protocol.core.methods.response.EthBlock$Block@3b0a4301,
          org.web3j.protocol.core.methods.response.EthBlock$Block@c8a60cff,
          org.web3j.protocol.core.methods.response.EthBlock$Block@77ccbd9d,
          org.web3j.protocol.core.methods.response.EthBlock$Block@1969d6a4,
          org.web3j.protocol.core.methods.response.EthBlock$Block@6ac0f99c,
          null,
          null,
          null,
          null,
          null]
        at app//maru.app.MaruFollowerTest.checkValidatorAndFollowerBlocks$lambda$16(MaruFollowerTest.kt:299)
```

Filter94:
Another [failure](https://github.com/Consensys/maru/actions/runs/16195193835/job/45719880746)

fluentcrafter:
Another failure https://github.com/Consensys/maru/actions/runs/16813674814/job/47625045203?pr=282

```
MaruFollowerTest > Maru follower is able to import blocks after going down() FAILED
    java.util.ConcurrentModificationException
        at java.base/java.util.HashMap$HashIterator.nextNode(HashMap.java:1605)
[...]

Issue #114, opened 2025-05-23
Flaky test - MaruConsensusSwitchTest


jframe:
The test MaruConsensusSwitchTest is failing frequently. Particularly the "Maru is capable of switching from Delegated to QBFT consensus without block pauses" test case.

It is often failing on the withdrawals validation rule in the EL with these log errors
```
[WARN ] 2025-05-23 03:05:45.037 [MinerExecutor] WithdrawalsValidator$AllowedWithdrawals - withdrawals must not be null when Withdrawals are activated
    [INFO ] 2025-05-23 03:05:45.037 [MinerExecutor] MainnetBlockValidator - Invalid block 4 (0xdee7c8a5247901bd9f525c3f9439bca9c9a82ec30b8031a8ff5ad31ef786cc9a): Optional[failed to validate output of imported block]
```

Also seen it failing due to unexpected clique extra data.

Filter94:
Another case of failure: https://github.com/Consensys/maru/actions/runs/15489816158/job/43613300329?pr=140

fluentcrafter:
Another failure that seems related:

https://github.com/Consensys/maru/actions/runs/15981522953/job/45076776598

```
 [WARN ] 2025-06-30 19:09:38.907 [MinerExecutor] WithdrawalsValidator$AllowedWithdrawals - withdrawals must not be null when Withdrawals are activated
    [INFO ] 2025-06-30 19:09:38.908 [MinerExecutor] MainnetBlockValidator - Invalid block 1 (0x4b2b235af63c816132a19dce341bf07deb8b262f7fd11bf0dfbd4111bf1a9e8b): Optional[failed to validate output of imported block]
Error: ROR] 2025-06-30 19:09:38.908 [MinerExecutor] BlockMiner - Illegal block mined, could not be imported to local chain.
```

Filter94:
Another example https://github.com/Consensys/maru/actions/runs/16059150006/job/45320794136

fluentcrafter:
Another example with block times

```
MaruConsensusSwitchTest > Maru is capable of switching from Delegated to QBFT consensus without block pauses() FAILED
    java.lang.AssertionError: invalid block time: expected=1 actual=2 blocks timestamps=[1752143470, 1752143472]
        at maru.testutils.Checks.verifyBlockTime(Checks.kt:42)
        at maru.testutils.Checks.verifyBlockTimeWithAGapOn(Checks.kt:54)
[...]

PR #178, opened 2025-07-09
adjust testing parallel configs to aliviate CI issues

fluentcrafter:


fluentcrafter:
This should alleviate issues:
- #168 
- #114

PR #184, opened 2025-07-10
Parallel gradle build

fluentcrafter:


app/build.gradle
```diff
@@ -80,7 +94,47 @@ jar {
...
-test {
+task integrationTest(type: Test) {
+  description = "Runs integration tests."
+  group = "verification"
+  useJUnitPlatform()
+
+  classpath = sourceSets.integrationTest.runtimeClasspath
+  testClassesDirs = sourceSets.integrationTest.output.classesDirs
+
+  maxHeapSize = "4096m"
```
gauravahuja, line 105: Why specify the max heap size here, gradle.properties already specifies it and it is currently same as this value.

gauravahuja (approved):

gauravahuja (approved):

app/build.gradle
Filter94: Good question. It wasn't working for me without it initially, but now it works

PR #188, opened 2025-07-10
Flacky test fix 2

fluentcrafter:


app/src/test/kotlin/maru/app/MaruLineaFinalizationTest.kt
```diff
@@ -96,6 +97,18 @@ class MaruLineaFinalizationTest {
         requestRetryConfig = null,
         vertx = null,
       )
+    // wait for Besu to be fully started and synced, to avoid CI flakiness due low resources some times
+    await
+      .atMost(5.seconds.toJavaDuration())
+      .pollInterval(200.milliseconds.toJavaDuration())
+      .untilAsserted {
```
Filter94, line 104: Should we add `.ignoreExceptions()` here?

fluentcrafter: Definitely yes! Good call, thanks

Filter94 (approved):

Issue #121, opened 2025-05-28
P2PTest parallelization


Filter94:
## Description
`P2PTest` was developed with hardcoded set of used ports, which prevents this testsuite's parallelization

## Motivation
The tests would move faster, improving the Developer experience and saving the CI cost

## Tasks
- [ ] Change `P2PTest` setup to use vacant ports for the P2PNetwork so it's possible to run the tests in parallel

## Acceptance criteria
- [ ] `P2PTest` is running successfully in parallel, without flackyness

## Risks
- [ ] We could have unnoticed flackyness

## Remember to
 - [ ] Add the `documentation` label in case there is an impact on the documentation
 - [ ] Add `priority` and `team` labels
 - [ ] Add Task for updating the Runbook or adding/updating existing metrics and alerts.

PR #196, opened 2025-07-14
BeaconBlocksByRange message

jframe:
Implements the BeaconBlocksByRange message as resq/resp so it can be used to sync other nodes


core/src/main/kotlin/maru/database/BeaconChain.kt
```diff
@@ -24,6 +24,17 @@ interface BeaconChain : AutoCloseable {
 
   fun getSealedBeaconBlock(beaconBlockNumber: ULong): SealedBeaconBlock?
 
+  fun getSealedBlocks(
+    startBlockNumber: ULong,
+    count: ULong,
+  ): List<SealedBeaconBlock> =
+    generateSequence(startBlockNumber) { it + 1UL }
+      .take(count.toInt())
```
Filter94, line 32:
```suggestion
    (startBlockNumber..startBlockNumber + count)
```

core/src/test/kotlin/maru/core/InMemoryBeaconChainTest.kt
```diff
@@ -139,4 +139,72 @@ class InMemoryBeaconChainTest {
...
+    testBlocks.forEach { block ->
+      updater.putSealedBeaconBlock(block)
+    }
+    updater.commit()
+
+    val blocks = inMemoryBeaconChain.getSealedBlocks(startBlockNumber = 2uL, count = 3uL)
+    assertThat(blocks).hasSize(3)
+    assertThat(blocks[0]).isEqualTo(testBlocks[1]) // block 2
+    assertThat(blocks[1]).isEqualTo(testBlocks[2]) // block 3
+    assertThat(blocks[2]).isEqualTo(testBlocks[3]) // block 4
```
Filter94, line 157: Maybe we can turn it into a sublist comparison?

core/src/test/kotlin/maru/core/InMemoryBeaconChainTest.kt
```diff
@@ -139,4 +139,72 @@ class InMemoryBeaconChainTest {
...
+
+    val updater = inMemoryBeaconChain.newUpdater()
+    updater.putSealedBeaconBlock(testBlock).commit()
+
+    val blocks = inMemoryBeaconChain.getSealedBlocks(startBlockNumber = 1uL, count = 0uL)
+    assertThat(blocks).isEmpty()
+  }
+
+  @Test
+  fun `getSealedBlocks stops at gap in sequence`() {
```
Filter94, line 178: Would it be better if it threw an exception in this case? I believe it shouldn't happen under normal circumstances

p2p/src/main/kotlin/maru/p2p/RpcMethods.kt
```diff
@@ -31,7 +37,25 @@ class RpcMethods(
     )
   }
 
+  val beaconBlocksByRangeRequestMessageSerDe = BeaconBlocksByRangeRequestMessageSerDe()
+  val beaconBlocksByRangeResponseMessageSerDe =
+    BeaconBlocksByRangeResponseMessageSerDe(RLPSerializers.SealedBeaconBlockSerializer)
+
+  val beaconBlocksByRangeRpcMethod by lazy {
```
Filter94, line 44: Why are we making it lazy?

p2p/src/test/kotlin/maru/p2p/BeaconBlocksByRangeIntegrationTest.kt
```diff
@@ -0,0 +1,285 @@
...
+import org.junit.jupiter.api.parallel.ExecutionMode
+import tech.pegasys.teku.networking.p2p.libp2p.LibP2PNodeId
+
+@Execution(ExecutionMode.SAME_THREAD)
+class BeaconBlocksByRangeIntegrationTest {
+  companion object {
+    private val chainId = 1337u
+    private const val IPV4 = "127.0.0.1"
+    private const val PORT1 = 9334u
+    private const val PORT2 = 9335u
```
Filter94, line 48: The more we're delaying dynamic port allocation in the tests, the more this approach is copy pasted and  more tech debt we introduce. Maybe we should prioritize [this](https://github.com/Consensys/maru/issues/121) or at least develop new, parallelizable approach for the new tests we introduce?

p2p/src/test/kotlin/maru/p2p/BeaconBlocksByRangeIntegrationTest.kt
```diff
@@ -0,0 +1,285 @@
...
+        payload = request,
+      )
+
+    val responseHandler = MaruRpcResponseHandler<Message<BeaconBlocksByRangeResponse, RpcMessageType>>()
+
+    peer1.sendRequest(
+      createRpcMethods(beaconChain2).beaconBlocksByRange(),
+      requestMessage,
+      responseHandler,
+    )
```
Filter94, line 231: I think this part is too low level to be tested here and to be exposed to the user's code. Maybe we should instead instantiate all the necessary stuff with the same factory that will be used by Maru itself?

And will request be sent with `peer1.sendRequest` or there will be more high-level interface to interact with?

p2p/src/test/kotlin/maru/p2p/BeaconBlocksByRangeIntegrationTest.kt
```diff
@@ -0,0 +1,285 @@
...
+
+    val responseMessage = responseHandler.response().get(5, TimeUnit.SECONDS)
+    val response = responseMessage.payload
+
+    // Verify empty response
+    assertThat(response.blocks).isEmpty()
+  }
+
+  private fun createForkIdHashProvider(beaconChain: BeaconChain): ForkIdHashProvider {
```
Filter94, line 240: This is also too low level IMO. Especially since the logic to create ForkIdHashProvider is there already in `MaruAppFactory`. I think we should factor out a factory class to instantiate P2PNetwork and use it here

p2p/src/test/kotlin/maru/p2p/messages/BeaconBlocksByRangeHandlerTest.kt
```diff
@@ -0,0 +1,123 @@
...
+
+class BeaconBlocksByRangeHandlerTest {
+  private lateinit var beaconChain: BeaconChain
+  private lateinit var handler: BeaconBlocksByRangeHandler
+  private lateinit var peer: MaruPeer
+  private lateinit var callback: ResponseCallback<Message<BeaconBlocksByRangeResponse, RpcMessageType>>
+
+  @BeforeEach
+  fun setup() {
+    beaconChain = mock()
```
Filter94, line 35: Is mock really much easier to use here than the `InMemoryBeaconChain`?

storage/src/test/kotlin/maru/database/kv/KvDatabaseTest.kt
```diff
@@ -179,4 +179,100 @@ class KvDatabaseTest {
       assertThat(db.getSealedBeaconBlock(testBeaconBlock2Number)).isNull()
     }
   }
+
+  @Test
```
Filter94, line 183: I created a ticket to use single testsuite for InMemory and the persistent implementations https://github.com/Consensys/maru/issues/198

core/src/test/kotlin/maru/core/InMemoryBeaconChainTest.kt
jframe: Agree, changed to throw an exception

core/src/main/kotlin/maru/database/BeaconChain.kt
jframe: Tried that but it breaks returning no blocks when the count is zero. Could be fixed with some use of max but I think it ends cleaner this way keeping the take with use of max and min.

p2p/src/test/kotlin/maru/p2p/P2PTest.kt
```diff
@@ -432,6 +462,77 @@ class P2PTest {
...
+      p2PNetworkImpl1.start()
+      p2pManagerImpl2.start()
+
+      awaitUntilAsserted { assertNetworkHasPeers(network = p2PNetworkImpl1, peers = 1) }
+      awaitUntilAsserted { assertNetworkHasPeers(network = p2pManagerImpl2, peers = 1) }
+
+      val peer1 =
+        p2pManagerImpl2.peerLookup.getPeer(LibP2PNodeId(PeerId.fromBase58(PEER_ID_NODE_1)))
+          ?: throw IllegalStateException("Peer with ID $PEER_ID_NODE_1 not found in p2pManagerImpl2")
+      val maruPeer1 = DefaultMaruPeer(peer1, rpcMethods, statusMessageFactory)
```
Filter94, line 519: peer1 is already is a `DefaultMaruPeer` and I think this means that user of p2pManagerImpl2 instance doesn't need the `rpcMethods`. Same applies to the other usage of `rpcMethods`, meaning that we should be able to remove it

core/src/main/kotlin/maru/database/BeaconChain.kt
```diff
@@ -24,6 +25,17 @@ interface BeaconChain : AutoCloseable {
 
   fun getSealedBeaconBlock(beaconBlockNumber: ULong): SealedBeaconBlock?
 
+  fun getSealedBlocks(
```
gauravahuja, line 28: getSealedBeaconBlocks

core/src/main/kotlin/maru/database/BeaconChain.kt
```diff
@@ -24,6 +25,17 @@ interface BeaconChain : AutoCloseable {
...
+    startBlockNumber: ULong,
```
gauravahuja, line 29: add a comment or make the name more explicit about it being inclusive or exclusive start block number.

core/src/main/kotlin/maru/p2p/Message.kt
```diff
@@ -19,6 +19,7 @@ enum class GossipMessageType : MessageType {
 
 enum class RpcMessageType : MessageType {
   STATUS(),
+  BEACON_BLOCKS_BY_RANGE(),
```
gauravahuja, line 22: I don't think you need the braces here.

p2p/src/main/kotlin/maru/p2p/messages/BeaconBlocksByRangeRequest.kt
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Request message for fetching beacon blocks by range.
+ * Similar to BeaconBlocksByRange in Ethereum consensus specs but adapted for Maru.
+ *
+ * @param startBlockNumber The block number to start fetching from
+ * @param count The number of blocks to fetch (maximum)
+ */
+data class BeaconBlocksByRangeRequest(
+  val startBlockNumber: ULong,
```
gauravahuja, line 19: make it explicit, startBlockNumberInclusive or startBlockNumberExclusive

jframe: don't think that is necessary as Java APIs in general are inclusive for the start. I've also added javadoc to make it clear it's inclusive

p2p/src/test/kotlin/maru/p2p/P2PTest.kt
```diff
@@ -80,12 +82,25 @@ class P2PTest {
...
       val rpcProtocolIdGenerator = LineaRpcProtocolIdGenerator(chainId)
       lateinit var maruPeerManager: MaruPeerManager
-      val rpcMethods = RpcMethods(statusMessageFactory, rpcProtocolIdGenerator) { maruPeerManager }
+      val rpcMethods = RpcMethods(statusMessageFactory, rpcProtocolIdGenerator, { maruPeerManager }, beaconChain)
       val maruPeerFactory = DefaultMaruPeerFactory(rpcMethods, statusMessageFactory)
       maruPeerManager = MaruPeerManager(maruPeerFactory = maruPeerFactory)
       return rpcMethods
     }
 
+    fun createRpcMethods(
```
Filter94, line 91: But these 2 `createRpcMethods` aren't used now

Filter94 (approved):

gauravahuja (approved):

Issue #163, opened 2025-06-30
Spike for MetricsSystem adapter


gauravahuja:
Currently we are using two metrics framework in maru, 
- Besu's metrics system
- Linea's metrics facade

we do not want to setup two separate metrics servers on different ports for prometheus to scrape. So, we want to investigate a way to have an adapter for metrics system that uses metrics facade behind the scenes.

PR #202, opened 2025-07-15
[163] Implement adapter for Besu MetricsSystem using MetricsFacade

gauravahuja:
#163


observability/src/main/kotlin/maru/metrics/BesuMetricsSystemAdapter.kt
```diff
@@ -0,0 +1,151 @@
...
+    name: String,
+    help: String,
+    vararg labelNames: String,
+  ): LabelledMetric<OperationTimer> =
+    createLabelledTimer(
+      category = category,
+      name = name,
+      help = help,
+      labelNames = labelNames,
+    )
```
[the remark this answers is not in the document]
gauravahuja: If you look at the [implementation in Besu](https://github.com/hyperledger/besu/blob/main/metrics/core/src/main/java/org/hyperledger/besu/metrics/prometheus/PrometheusMetricsSystem.java#L144), the LabelledTimer has defined quantiles whereas SimpleLabelledTimer does not. However we do not have a concept of quantiles for timers in MetricsFacade so I cannot have an adapter that implements it.

observability/src/test/kotlin/maru/metrics/BesuMetricsSystemAdapterTest.kt
```diff
@@ -0,0 +1,251 @@
...
+  fun `creates counter with value supplier`() {
+    val expectedTags = listOf(Tag("key1", "value1"), Tag("key2", "value2"))
+    val labelledMetric =
+      besuMetricsSystemAdapter.createLabelledSuppliedCounter(
+        category = BesuMetricsCategoryAdapter.from(TestLineaCategory.TEST_CATEGORY),
+        name = "some.metric",
+        help = "This is a test counter",
+        labelNames = expectedTags.map { it.key }.toList().toTypedArray(),
+      )
+    val counter =
```
jonesho, line 231: chore: this line `val counter =` can be removed

observability/src/test/kotlin/maru/metrics/BesuMetricsSystemAdapterTest.kt
```diff
@@ -0,0 +1,251 @@
...
+        { 2.0 },
+        *expectedTags.map { it.value }.toList().toTypedArray(),
+      )
+    val createdCounter = meterRegistry.find("linea.test.test.category.some.metric").counter()
+    assertThat(createdCounter).isNotNull
+
+    await
+      .pollInterval(1.seconds.toJavaDuration())
+      .timeout(10.seconds.toJavaDuration())
+      .untilAsserted { createdCounter!!.count() >= 8.0 }
```
jonesho, line 242: I tried this locally and this is always asserted as true, pls change to
`.untilAsserted { assertThat(createdCounter!!.count()).isGreaterThanOrEqualTo(8.0) }`

gauravahuja: Oh, thanks for catching it. Fixed.

jframe (approved):

PR #203, opened 2025-07-15
Improv configs parsing

fluentcrafter:


config/src/main/kotlin/maru/config/ConfigLoaderHelper.kt
```diff
@@ -0,0 +1,121 @@
...
+  logger: Logger = LogManager.getLogger("maru.config"),
+  strict: Boolean,
+): Result<T, String> =
+  loadConfigsOrError<T>(configFiles, strict = strict)
+    .also {
+      val logLevel = if (strict) Level.WARN else Level.ERROR
+      logErrorIfPresent(it, logger, logLevel)
+    }
+
+inline fun <reified T : Any> loadConfigsOrError(
```
jonesho, line 91: Should this inline function be removed? Why don't we just call `loadConfigsAndLogErrors` instead? This also confuses me a bit with the overloading inline function `loadConfigsOrError` above, wdyt?

jonesho (approved):

jonesho (approved):

PR #204, opened 2025-07-16
Change beacon chain initialisation to use empty block for payload instead of the latest EL block

jframe:
This changes the BeaconChain initialisation to use an empty payload instead of the latest execution layer block in the genesis block.

Using the latest EL payload in the genesis has some issues for syncing
1. Each node could initialise the BeaconChain with a different EL block included in the genesis block since we are using the latest block. The EL might be behind when we get the latest block.
2. New nodes started after Maru genesis block is created will have a different EL block as the payload as they will use the latest block will be different.
3. It requires the EL to be fully sync before the CL is synced. After the merge the EL requires the CL to be in sync first. For the merge is probably ok, post merge new nodes may have trouble syncing.


app/src/main/kotlin/maru/app/MaruApp.kt
```diff
@@ -168,10 +169,17 @@ class MaruApp(
       NewBlockHandlerMultiplexer(followerBlockHandlers)
     val adaptedBeaconBlockImporter = SealedBeaconBlockHandlerAdapter(blockImportHandlers)
 
+    val qbftForkTimestamp =
```
Filter94, line 172: Maybe we should introduce this as a dedicated method rather than exposing `getAllForks`?

consensus/src/test/kotlin/maru/consensus/qbft/EagerQbftBlockCreatorTest.kt
```diff
@@ -225,14 +227,48 @@ class EagerQbftBlockCreatorTest {
...
+    // Spy on the ExecutionLayerManager to capture calls
+    val spyExecutionLayerManager = Mockito.spy(executionLayerManager)
+    val eagerQbftBlockCreator = setup(sealedGenesisBeaconBlock, parentHeader, spyExecutionLayerManager)
+
+    setNextSecondMillis(0)
+    val acceptedBlockTimestamp = (clock.millis() / 1000)
+    eagerQbftBlockCreator.createBlock(acceptedBlockTimestamp, parentHeader)
+
+    // Verify that getLatestBlockHash() was called since parent block number is 0 (genesis)
+    Mockito.verify(spyExecutionLayerManager).getLatestBlockHash()
```
Filter94, line 260: Can we instead verify the parent of the new block?

Filter94 (dismissed):

app/src/main/kotlin/maru/app/MaruApp.kt
jframe: Yeah good idea, it did feel a bit wrong exposing that method. I'll add that

consensus/src/test/kotlin/maru/consensus/qbft/EagerQbftBlockCreatorTest.kt
jframe: No unfortunately as the parent is same in both cases. Need to verify that the hash it's using the build the block is coming from the executionLayerManager instead of the block executionPayload hash. I've improved the test so that the beacon parent block doesn't have the execution layer hash in it now as well.

Filter94: I think what we really want to verify is that it's able to produce a valid block when `parentHeader` contains an empty Genesis block.

Your test is focused on _how_ rather than _what_, which makes it fragile. In the test shouldn't care _how_ it didn't get stuck on a genesis block, we should just make sure it produces something

PR #205, opened 2025-07-16
Configs consistency

fluentcrafter:


config/src/test/kotlin/maru/config/HopliteFriendlinessTest.kt
```diff
@@ -59,7 +59,10 @@ class HopliteFriendlinessTest {
     P2P(
       ipAddress = "127.0.0.1",
       port = 3322u,
-      staticPeers = emptyList(),
+      staticPeers =
+        listOf(
+          "/ip4/maru-validator-0.maru-validator.default.svc.cluster.local/tcp/3322/p2p/16Uiu2HAmFjVuJoKD6sobrxwyJyysM1rgCsfWKzFLwvdB2HKuHwTg",
```
Filter94, line 64: Nit: This is a bit misleading example. While it doesn't really matter for this test, can we still put a valid Multiaddr here? Either `ip4` or `dns4`?

fluentcrafter: done, thanks

Issue #214, opened 2025-07-18
Spike/Interfaces for Syncing


gauravahuja:
Preliminary spike with interfaces and flows for syncing mechanism in maru. On the lines of the synch process mentioned in this [document](https://docs.google.com/document/d/1tBGpr3i-G5N4OMAP-xMWtDjoirfVul4b50Ugqy2rfA4/edit?usp=sharing)

PR #216, opened 2025-07-21
Interfaces for Syncing

gauravahuja:
#214 

Based on the Sync Flow described in [this chart](https://drive.google.com/file/d/1kzQRuDZCsW7bCMFET3iBkQEosNN1gn3S/view?usp=sharing)


syncing/src/main/kotlin/maru/syncing/SyncController.kt
```diff
@@ -0,0 +1,23 @@
...
+
+  fun stop()
+
+  fun isBeaconChainSynced(): Boolean
+
+  fun isELSynced(): Boolean
+
+  fun onBeaconSyncComplete(handler: () -> Unit)
+
+  fun onELSyncComplete(handler: () -> Unit)
```
Filter94, line 22: I feel like it's wrong to put these methods here. IMO they belong to s separate SyncStatus class, which would be responsible for the state management and access (isBeaconChainSynced, isELSynced). Reg. subscription methods, I'm less clear, whether they should be moved somewhere or not

syncing/src/main/kotlin/maru/syncing/ChainTracker.kt
```diff
@@ -0,0 +1,18 @@
...
+ */
+package maru.syncing
+import maru.core.SealedBeaconBlock
+
+interface ChainTracker {
+  fun onChainHeadUpdated(chainHeadUpdatedHandler: ChainHeadUpdatedHandler)
+}
+
+interface ChainHeadUpdatedHandler {
+  fun onChainHeadUpdated(headSealedBeaconBlock: SealedBeaconBlock)
```
Filter94, line 17: Since we're going to rely on status messages for determining the chain head, I don't think `SealedBeaconBlock` should be here. I think it has to be `BeaconSyncTarget`

PR #231, opened 2025-07-23
Ignoring errors returned during new block import by the follower nodes

Filter94:
Since we don't guarantee high consistency for the Follower nodes, the logic should be fire and forget. Otherwise, currently it will cause Maru to get stuck in case any of the follower nodes fails to import new block


fluentcrafter (approved):

Filter94:
I'm not sure if we need it. I'll convert it to a draft for now and will probably close it in the future

Issue #221, opened 2025-07-22
Implement PeerChainTracker


Filter94:
## Description
We need a component that would gather chain heads reported by peers. For 5 peers it should have 5 records with Peer's latest known chain state (chain height). This component should expose the whole structure as a public function.

[Teku code](https://github.com/Consensys/teku/blob/master/beacon/sync/src/main/java/tech/pegasys/teku/beacon/sync/forward/multipeer/chains/PeerChainTracker.java) for inspiration

## Motivation
We need to maintain the up to date chain state of our peers to implement `SyncTargetSelector`

## Tasks
- [ ] Implement the data collection to keep 1 latest record per peer
- [ ] Implement data pruning. Old records have to be deleted
- [ ] Implement PeersHeadBlockProvider, a wrapper over MaruPeerManager

## Acceptance criteria
- [ ] `PeerChainTracker` tracks the last known peer status
- [ ] `PeerChainTracker` doesn't retain stale peers indefinitely

## Risks
- [ ] If peers are retained indefinitely, it may be considered as a minor memory leak

## Remember to
 - [ ] Add the `documentation` label in case there is an impact on the documentation
 - [ ] Add `priority` and `team` labels
 - [ ] Add Task for updating the Runbook or adding/updating existing metrics and alerts.

PR #241, opened 2025-07-25
Implemented P2PPeersHeadBlockProvider

Filter94:
Implements #221

### DIFF
--- a/p2p/src/main/kotlin/maru/p2p/P2PPeersHeadBlockProvider.kt
+++ b/p2p/src/main/kotlin/maru/p2p/P2PPeersHeadBlockProvider.kt
@@ -1,0 +1,39 @@
+/*
+ * Copyright Consensys Software Inc.
+ *
+ * This file is dual-licensed under either the MIT license or Apache License 2.0.
+ * See the LICENSE-MIT and LICENSE-APACHE files in the repository root for details.
+ *
+ * SPDX-License-Identifier: MIT OR Apache-2.0
+ */
+package maru.p2p
+
+import org.apache.logging.log4j.LogManager
+import org.apache.logging.log4j.Logger
+
+class P2PPeersHeadBlockProvider(
+  private val peerLookup: PeerLookup,
+) : PeersHeadBlockProvider {
+  private val log: Logger = LogManager.getLogger(P2PPeersHeadBlockProvider::class.java)
+
+  /**
+   * Returns a map of peer IDs to their latest block numbers.
+   * Only includes peers that have a valid status with a block number.
+   */
+  override fun getPeersHeads(): Map<String, ULong> =
+    peerLookup
+      .getPeers()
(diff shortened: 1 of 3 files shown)


p2p/src/main/kotlin/maru/p2p/P2PPeersHeadBlockProvider.kt
```diff
@@ -0,0 +1,39 @@
...
+  private val log: Logger = LogManager.getLogger(P2PPeersHeadBlockProvider::class.java)
+
+  /**
+   * Returns a map of peer IDs to their latest block numbers.
+   * Only includes peers that have a valid status with a block number.
+   */
+  override fun getPeersHeads(): Map<String, ULong> =
+    peerLookup
+      .getPeers()
+      .mapNotNull { peer ->
```
gauravahuja, line 26: This will make all the requests sequential when they don't have to be. I don't know how many peers we will have but if we have to query 100s or 1000s or peers it will take long time to do it sequentially or without limiting the number of peers.

Filter94: There are no requests here. `PeerLookup` provides the cached view of peer statuses

gauravahuja: oh wait, so from what I see in [MaruPeerManager](https://github.com/Consensys/maru/blob/main/p2p/src/main/kotlin/maru/p2p/MaruPeerManager.kt#L116), the status is received from the peer only once when the connection is established, it is not periodically updated?

Filter94: Right now - no, because in the initial implementation it wasn't clear, what do we want from statuses. But I thought it should be a responsibility of the P2PNetwork to update the status periodically. @jframe Do you have the same understanding?

jframe: Using the `peerLookup.getPeers()` will work correctly once that is implemented.

p2p/src/main/kotlin/maru/p2p/P2PPeersHeadBlockProvider.kt
```diff
@@ -0,0 +1,39 @@
...
+        try {
+          val status = peer.getStatus()
+          if (status != null) {
+            peer.id.toBase58() to status.latestBlockNumber
```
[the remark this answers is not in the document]
Filter94: I can make things slightly better by using `peer.toPeerInfo().nodeId`, but in general, I feel like `MaruPeer` is messy in general in terms of testability and separation of concerns. But this spans beyond the scope of this PR

fluentcrafter (approved):

PR #242, opened 2025-07-28
improve the download block step

pinges:
This adds the following to the download block step:

- read full range of blocks when only a part of the blocks are returned by peers
- timeout block request after 5s
- retry after
  - empty response
  - timeout
  - other exceptions
- throw exception after too many retries


syncing/src/main/kotlin/maru/syncing/pipeline/DownloadCompleteBlockRangeTask.kt
```diff
@@ -0,0 +1,68 @@
...
+      var startBlockNumber = targetRange.startBlock
+      var count = targetRange.endBlock - targetRange.startBlock + 1uL
+      var remaining = count
+      val downloadedBlocks = mutableListOf<SealedBeaconBlock>()
+      var retries = 0
+      do {
+        val peer = peerLookup.getPeers().random() // TODO: filter peers? Do we want to use least busy peer?
+        try {
+          peer
+            .sendBeaconBlocksByRange(startBlockNumber, remaining)
```
Filter94, line 35: How is `remaining` supposed to work in practice? What if the number of blocks doesn't fit into the recipient's memory, but the sender won't mind sending massive number of blocks over the network?

Or to put the question differently, aren't we supposed to throttle the count with some upper bound on the client side?

syncing/src/main/kotlin/maru/syncing/pipeline/DownloadCompleteBlockRangeTask.kt
```diff
@@ -0,0 +1,68 @@
...
+            .orTimeout(5L, TimeUnit.SECONDS)
+            .thenApply { response ->
+              if (response.blocks.isEmpty()) {
+                peer.adjustReputation(ReputationAdjustment.SMALL_PENALTY)
+                log.debug("No blocks received from peer: {}", peer.id)
+                retries++
+              } else {
+                // TODO: can we check if the blocks are valid?
+                // we could check the first block number, and whether the following block numbers increment by 1, as well as the parent roots
```
Filter94, line 44: An alternative could be to return blocks with the peer object, so we could penalize peers later on

syncing/src/main/kotlin/maru/syncing/pipeline/DownloadCompleteBlockRangeTask.kt
```diff
@@ -0,0 +1,68 @@
...
+        } catch (e: Exception) {
+          if (e.cause is TimeoutException) {
+            log.debug("Timed out while downloading blocks from peer: {}", peer.id)
+            peer.adjustReputation(ReputationAdjustment.LARGE_PENALTY)
+          } else {
+            log.debug("Failed to download blocks from peer: {}", peer.id, e)
+          }
+          retries++
+        }
+        if (retries > 5) {
```
Filter94, line 60: This should definitely be configurable

syncing/src/main/kotlin/maru/syncing/pipeline/DownloadCompleteBlockRangeTask.kt
```diff
@@ -0,0 +1,68 @@
...
+      var count = targetRange.endBlock - targetRange.startBlock + 1uL
+      var remaining = count
+      val downloadedBlocks = mutableListOf<SealedBeaconBlock>()
+      var retries = 0
+      do {
+        val peer = peerLookup.getPeers().random() // TODO: filter peers? Do we want to use least busy peer?
+        try {
+          peer
+            .sendBeaconBlocksByRange(startBlockNumber, remaining)
+            .orTimeout(5L, TimeUnit.SECONDS)
```
Filter94, line 36: This has to be configurable

syncing/src/test/kotlin/maru/syncing/pipeline/BeaconChainDownloadPipelineFactoryTest.kt
```diff
@@ -183,20 +185,23 @@ class BeaconChainDownloadPipelineFactoryTest {
...
 
+    val block1 = randomSealedBeaconBlock(ULong.MAX_VALUE - 2uL)
+    val block2 = randomSealedBeaconBlock(ULong.MAX_VALUE - 1uL)
+
     // Test with a range very close to ULong.MAX_VALUE
-    val startBlock = ULong.MAX_VALUE - 20uL
+    val startBlock = ULong.MAX_VALUE - 2uL
     val endBlock = ULong.MAX_VALUE - 1uL
 
     // The expected ranges with request size 10
```
Filter94, line 195: This is now wrong

syncing/src/test/kotlin/maru/syncing/pipeline/DownloadBlocksStepTest.kt
```diff
@@ -35,21 +35,6 @@ class DownloadBlocksStepTest {
     assertThat(result).isEqualTo(blocks)
   }
 
-  @Test
-  fun `returns empty list if peer returns empty response`() {
```
Filter94, line 39: Why is this test irrelevant anymore?

syncing/src/main/kotlin/maru/syncing/pipeline/DownloadCompleteBlockRangeTask.kt
pinges: I was thinking the same thing. Because only when we are able to import the block we really know whether it was a valid block or not. And we should be able to penalise/disconnect the peer at that point.

syncing/src/main/kotlin/maru/syncing/pipeline/DownloadCompleteBlockRangeTask.kt
```diff
@@ -0,0 +1,69 @@
...
+            .thenApply { response ->
+              if (response.blocks.isEmpty()) {
+                peer.adjustReputation(ReputationAdjustment.SMALL_PENALTY)
+                log.debug("No blocks received from peer: {}", peer.id)
+                retries++
```
jframe, line 41: Should probably wait before retrying as well

syncing/src/main/kotlin/maru/syncing/pipeline/DownloadCompleteBlockRangeTask.kt
```diff
@@ -0,0 +1,69 @@
...
+              } else {
+                // TODO: can we check if the blocks are valid?
+                // we could check the first block number, and whether the following block numbers increment by 1, as well as the parent roots
```
jframe, line 44: Agree that at the very least we should verify these are the blocks we asked for i.e. they have the block numbers we expected. That way we can retry another peer.

syncing/src/main/kotlin/maru/syncing/pipeline/DownloadCompleteBlockRangeTask.kt
```diff
@@ -0,0 +1,68 @@
...
+
+  fun getCompleteBlockRange(targetRange: SyncTargetRange): CompletableFuture<List<SealedBeaconBlock>> =
+    CompletableFuture<List<SealedBeaconBlock>>.supplyAsync {
+      var startBlockNumber = targetRange.startBlock
+      var count = targetRange.endBlock - targetRange.startBlock + 1uL
+      var remaining = count
+      val downloadedBlocks = mutableListOf<SealedBeaconBlock>()
+      var retries = 0
+      do {
+        val peer = peerLookup.getPeers().random() // TODO: filter peers? Do we want to use least busy peer?
```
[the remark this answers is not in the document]
pinges: Future PR?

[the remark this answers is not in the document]
Filter94: Yes, I think it's an improvement, rather than a blocking issue

syncing/src/main/kotlin/maru/syncing/pipeline/DownloadCompleteBlockRangeTask.kt
pinges: There is a max defined for the number of blocks that you can request.
I guess we do select a default number of blocks that we request at a time, which cannot be greater than  max.
We should implement throttling in another PR.

syncing/src/main/kotlin/maru/syncing/pipeline/DownloadCompleteBlockRangeTask.kt
```diff
@@ -0,0 +1,68 @@
...
+                retries++
+              } else {
+                // TODO: can we check if the blocks are valid?
+                // we could check the first block number, and whether the following block numbers increment by 1, as well as the parent roots
+                downloadedBlocks.addAll(response.blocks)
+                val numBlocks = response.blocks.size.toULong()
+                startBlockNumber += numBlocks
+                remaining -= numBlocks
+              }
+            }.join()
```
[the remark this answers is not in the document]
pinges: This could be done, by doing recursion, but we would have problems with the timeout the way it works at the moment. We could probably move the timeout inside the "sendBeaconBlocksByRange(startBlockNumber, remaining)" method ?
And I think doing that would make the code less readable as well.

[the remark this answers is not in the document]
Filter94: > And I think doing that would make the code less readable as well

Not sure about that. Shouldn't the timeout be on the client level in any case?

syncing/src/main/kotlin/maru/syncing/pipeline/DownloadCompleteBlockRangeTask.kt
```diff
@@ -0,0 +1,68 @@
...
+import java.util.concurrent.TimeoutException
+import maru.core.SealedBeaconBlock
+import maru.p2p.PeerLookup
+import org.apache.logging.log4j.LogManager
+import tech.pegasys.teku.networking.p2p.reputation.ReputationAdjustment
+
+class DownloadCompleteBlockRangeTask(
+  private val peerLookup: PeerLookup,
+) {
+  private val log: org.apache.logging.log4j.Logger = LogManager.getLogger(this::javaClass)
```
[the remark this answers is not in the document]
pinges: done

config/src/main/kotlin/maru/config/Config.kt
```diff
@@ -50,6 +50,13 @@ data class P2P(
   )
 }
 
+data class Sync(
```
Filter94, line 53: There's already `SyncingConfig` that I think we should use. But in your case it probably makes sense to introduce a subdivision under the `[syncing]` section, something like `[syncing.downloader]` or `[syncing.block-fetcher]`

syncing/src/main/kotlin/maru/syncing/pipeline/BeaconChainDownloadPipelineFactory.kt
```diff
@@ -23,7 +25,9 @@ class BeaconChainDownloadPipelineFactory(
   private val metricsSystem: MetricsSystem,
   private val peerLookup: PeerLookup,
   private val downloaderParallelism: Int,
```
Filter94, line 27: Usually the pattern we use to configure classes is to define a `data class Config`, which makes it easier to instantiate and refactor later on. In the client code it would be referred to as `BeaconChainDownloadPipelineFactory.Config`

syncing/src/main/kotlin/maru/syncing/pipeline/DownloadCompleteBlockRangeTask.kt
```diff
@@ -0,0 +1,86 @@
...
+
+class DownloadCompleteBlockRangeTask(
+  private val peerLookup: PeerLookup,
+  private val maxRetries: UInt,
+  private val blockRangeRequestTimeout: Duration,
+) : Function<SyncTargetRange, CompletableFuture<List<SealedBlockWithPeer>>> {
+  private val log: Logger = LogManager.getLogger(this::javaClass)
+
+  override fun apply(targetRange: SyncTargetRange): CompletableFuture<List<SealedBlockWithPeer>> =
+    CompletableFuture<List<SealedBeaconBlock>>.supplyAsync {
```
Filter94, line 32:
```suggestion
    CompletableFuture.supplyAsync {
```

I don't think type specification is any useful here

syncing/src/main/kotlin/maru/syncing/pipeline/DownloadCompleteBlockRangeTask.kt
```diff
@@ -0,0 +1,86 @@
...
+      var startBlockNumber = targetRange.startBlock
+      val count = targetRange.endBlock - targetRange.startBlock + 1uL
+      var remaining = count
+      val downloadedBlocks = mutableListOf<SealedBlockWithPeer>()
+      var retries = 0u
+      var peer: MaruPeer? = null
```
Filter94, line 38:
```suggestion
      var peer: MaruPeer?
```

Filter94 (dismissed):

config/src/main/kotlin/maru/config/Config.kt
pinges: like that idea

syncing/src/main/kotlin/maru/syncing/pipeline/BeaconChainDownloadPipelineFactory.kt
pinges: done

syncing/src/main/kotlin/maru/syncing/pipeline/DownloadCompleteBlockRangeTask.kt
pinges: done

jframe (approved):

PR #246, opened 2025-07-29
improv error logging

fluentcrafter:


Filter94 (commented):
It looks like one of our tests was relying on these error logs. Can you please fix it or relax the verifyable condition?

Filter94 (approved):

PR #251, opened 2025-07-30
Implement beacon chain syncing service

### LINKED ISSUE #227: Implement Beacon chain syncing service

jframe:
Implements to the beacon chain syncing service

fixes #227


syncing/src/main/kotlin/maru/syncing/beaconchain/CLSyncService.kt
```diff
@@ -0,0 +1,125 @@
...
+      name = "beaconchain.restart.counter",
+      description = "Count of chain pipeline restarts",
+    )
+
+  override fun setSyncTarget(syncTarget: ULong) {
+    log.info("Syncing started syncTarget={}", syncTarget)
+    this.syncTarget.set(syncTarget)
+
+    // If the pipeline is already running, we don't need to start a new one
+    if (pipeline == null) {
```
fluentcrafter, line 86: This check is also done inside `startSync()`, so maybe we can remove this one?

Also, reading startSycn code it seems that this check should be done only inside that function.

Given the check inside syncTarget, `setSyncTarget` should be `@Synchronised` method, otherwise if 2 different threads call it we may end up with 2 pipelines

syncing/src/test/kotlin/maru/syncing/beaconchain/CLSyncServiceImplTest.kt
```diff
@@ -0,0 +1,424 @@
...
+    }
+  }
+
+  @Test
+  fun `sync service download restarts on errors`() {
+    val signatureAlgorithm = SignatureAlgorithmFactory.getInstance()
+    val keypair = signatureAlgorithm.generateKeyPair()
+    val validator = Validator(Util.publicKeyToAddress(keypair.publicKey).toArray())
+    val validators = setOf(validator)
+    val genesisTimestamp = DataGenerators.randomTimestamp()
```
fluentcrafter, line 217: These seem C&P setup with above test. 

To improve test's readability, shall we place duplicated setup into shared class instance vars?

Also, if suitable leverage `@BeforeEach` notation for common setup taskes

syncing/src/main/kotlin/maru/syncing/beaconchain/CLSyncService.kt
```diff
@@ -0,0 +1,125 @@
...
+      }
+    }
+  }
+
+  override fun onSyncComplete(handler: (ULong) -> Unit) {
+    syncCompleteHanders.add(handler)
+  }
+
+  override fun start() {
+    executorService = Executors.newCachedThreadPool()
```
fluentcrafter, line 118: IIUC, this service has an -- not obvious -- implicit call order: 1 `start()`, 2 `setTarget()`

But if we invert the order, it may break because in L98 will call `pipeline.start(executorService)`  and the executor service may be null. 

IMO we should allow any call order to happen...

syncing/src/main/kotlin/maru/syncing/SyncController.kt
```diff
@@ -103,7 +110,17 @@ class SyncControllerImpl(
           executionLayerManager = elManager,
           onStatusChange = controller::elSyncStatusWasUpdated,
         )
-      val clSyncPipeline = CLSyncPipelineImpl()
+      val clSyncPipeline =
+        CLSyncServiceImpl(
+          beaconChain = beaconChain,
+          validators = validators,
+          allowEmptyBlocks = true,
```
fluentcrafter, line 117: `allowEmptyBlocks` should be an argument of the create Function, defaulted to false eventually

syncing/src/main/kotlin/maru/syncing/CLSyncService.kt
```diff
@@ -1,41 +0,0 @@
...
-
-interface CLSyncService {
-  fun setSyncTarget(syncTarget: ULong)
-
-  /**
-   * Notifies the handler when the <b>latest<b/> target is reached.
-   * If target is updated, onSyncComplete won't be called for previous targets
-   */
-  fun onSyncComplete(handler: (syncTarget: ULong) -> Unit)
-}
```
fluentcrafter, line 21: IMO we should keep the generic interfaces here in `maru.synching` package.

sync/src/test/kotlin/maru/sync/SyncServiceIntegrationTest.kt
```diff
@@ -0,0 +1,393 @@
...
+  private lateinit var p2pNetwork1: P2PNetworkImpl
+  private lateinit var p2pNetwork2: P2PNetworkImpl
+  private lateinit var forkIdHashProvider: ForkIdHashProvider
+  private val genesisState = DataGenerators.randomBeaconState(0u)
+
+  companion object {
+    private const val CHAIN_ID = 1337u
+    private const val IPV4 = "127.0.0.1"
+    private const val PORT1 = 9245u
+    private const val PORT2 = 9246u
```
fluentcrafter, line 56: Can we avoid using hardcoded ports? This is a source of "flacky" tests. We should instantiate the P2P on port 0, get a free port allocated and then build the PEER_ADDRESS_NODE based on that inside the @BeforeEach

Is that possible?

sync/src/main/kotlin/maru/sync/SyncService.kt
```diff
@@ -0,0 +1,119 @@
...
+import maru.core.Validator
+import maru.database.BeaconChain
+import maru.p2p.PeerLookup
+import maru.sync.pipeline.BeaconChainDownloadPipelineFactory
+import org.apache.logging.log4j.LogManager
+import org.apache.logging.log4j.Logger
+import org.hyperledger.besu.services.pipeline.Pipeline
+import tech.pegasys.teku.infrastructure.async.SafeFuture
+
+class SyncService(
```
fluentcrafter, line 39: Maybe I am lacking context, but I can't understand why we have 2 sync services: `CLSyncService` and `SyncService`  and what the responsibilityes of each?

syncing/src/test/kotlin/maru/syncing/beaconchain/CLSyncServiceImplTest.kt
```diff
@@ -0,0 +1,424 @@
...
+
+@Execution(ExecutionMode.SAME_THREAD)
+class CLSyncServiceImplTest {
+  companion object {
+    private val chainId = 1337u
+
+    private const val IPV4 = "127.0.0.1"
+
+    private const val PORT1 = 9234u
+    private const val PORT2 = 9235u
```
fluentcrafter, line 68: we should avoid hardcoded ports, it leads to flacky tests and we already have quite a few ... :(

syncing/src/test/kotlin/maru/syncing/beaconchain/CLSyncServiceImplTest.kt
```diff
@@ -0,0 +1,424 @@
...
+          validators = validators,
+          peerLookup = p2PNetworkImpl1.getPeerLookup(),
+          besuMetrics = TestMetricsSystemAdapter,
+          metricsFacade = TestMetricsFacade,
+          requestSize = 10u,
+          downloaderParallelism = 1u,
+        )
+      clSyncServiceImpl1.start()
+
+      var synced = false
```
fluentcrafter, line 122: this should be an atomic boolean because is being written and read by different threads...

Filter94 (commented):
Some initial feedback. I didn't go through the tests yet

sync/src/main/kotlin/maru/sync/SyncService.kt
Filter94: I also don't get this. Is the idea to start early with the testing with the real components and then replace `SyncService` with `CLSyncServiceImpl`?

sync/src/test/kotlin/maru/sync/SyncServiceIntegrationTest.kt
gauravahuja: Snippet to find a free port at runtime https://github.com/Consensys/linea-tracer/blob/arith-dev/testing/src/main/java/net/consensys/linea/testing/BesuExecutionTools.java#L335

consensus/src/main/kotlin/maru/consensus/qbft/DelayedQbftBlockCreator.kt
```diff
@@ -64,6 +65,37 @@ class DelayedQbftBlockCreator(
         )
       return QbftSealedBlockAdapter(sealedBlockBody)
     }
+
+    fun createBeaconBlock(
```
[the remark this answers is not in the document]
Filter94: Visible for testing == internal

syncing/src/test/kotlin/maru/syncing/beaconchain/CLSyncServiceImplTest.kt
```diff
@@ -0,0 +1,493 @@
...
+  @AfterEach
+  fun tearDown() {
+    p2pNetwork1.stop()
+    p2pNetwork2.stop()
+  }
+
+  @Test
+  fun `chain sync downloads and imports blocks from another node`() {
+    val synced = AtomicBoolean(false)
+    clSyncService.setSyncTarget(100uL)
```
Filter94, line 152: We should either name the magic number `100uL` and define it as a class field for reuse or you can just take the `beaconChain2.getLatestBeaconState().latestBeaconBlockHeader.number` with a proper name like `expectedChainHead` and use it

syncing/src/test/kotlin/maru/syncing/beaconchain/CLSyncServiceImplTest.kt
```diff
@@ -0,0 +1,493 @@
...
+    awaitUntilAsserted {
+      assertThat(handler1Called).isTrue()
+      assertThat(handler2Called).isTrue()
+    }
+  }
+
+  private fun createPeerAddress(port: UInt): MultiaddrPeerAddress =
+    MultiaddrPeerAddress.fromAddress("/ip4/$IPV4/tcp/$port/p2p/$PEER_ID_NODE_2")
+
+  private fun genesisState(
```
Filter94, line 348: Should we define this somewhere in the production code and reuse it here?

syncing/src/test/kotlin/maru/syncing/beaconchain/CLSyncServiceImplTest.kt
```diff
@@ -0,0 +1,493 @@
...
+  @BeforeEach
+  fun setUp() {
+    signatureAlgorithm = SignatureAlgorithmFactory.getInstance()
+    keypair = signatureAlgorithm.generateKeyPair()
+    validators = setOf(Validator(Util.publicKeyToAddress(keypair.publicKey).toArray()))
+
+    val genesisTimestamp = DataGenerators.randomTimestamp()
+    val (genesisBeaconState, genesisBeaconBlock) = genesisState(genesisTimestamp, validators)
+    beaconChain1 = spy(InMemoryBeaconChain(genesisBeaconState, genesisBeaconBlock))
+    beaconChain2 = spy(InMemoryBeaconChain(genesisBeaconState, genesisBeaconBlock))
```
Filter94, line 101: I would call it the `sourceBeaconChain`. The `beaconChain1` is then the the target

syncing/src/test/kotlin/maru/syncing/beaconchain/CLSyncServiceImplTest.kt
```diff
@@ -0,0 +1,493 @@
...
+    assertThat(beaconChain1.getLatestBeaconState()).isEqualTo(beaconChain2.getBeaconState(50uL))
+    for (i in 1uL..50uL) {
+      assertThat(beaconChain1.getSealedBeaconBlock(i)).isEqualTo(beaconChain2.getSealedBeaconBlock(i))
+      assertThat(beaconChain1.getBeaconState(i)).isEqualTo(beaconChain2.getBeaconState(i))
+    }
+
+    // update sync target to 100
+    synced.set(false)
+    clSyncService.setSyncTarget(100uL)
+    awaitUntilAsserted { assertThat(synced).isTrue() }
```
Filter94, line 181: I feel like these 3 are always together and could be factored out and reused. And maybe even the `val synced = AtomicBoolean(false)` can be defined inside this new function

syncing/src/test/kotlin/maru/syncing/beaconchain/CLSyncServiceImplTest.kt
```diff
@@ -0,0 +1,493 @@
...
+
+    assertThat(beaconChain1.getLatestBeaconState().latestBeaconBlockHeader.number).isEqualTo(100uL)
```
Filter94, line 183: I feel like assertions could also be factored out and reused

syncing/src/test/kotlin/maru/syncing/beaconchain/CLSyncServiceImplTest.kt
```diff
@@ -0,0 +1,493 @@
...
+import org.mockito.kotlin.any
+import org.mockito.kotlin.doAnswer
+import org.mockito.kotlin.never
+import org.mockito.kotlin.reset
+import org.mockito.kotlin.spy
+import org.mockito.kotlin.verify
+import org.mockito.kotlin.whenever
+import tech.pegasys.teku.networking.p2p.libp2p.MultiaddrPeerAddress
+
+class CLSyncServiceImplTest {
```
Filter94, line 68: We're not using an argument to onSyncComplete handler anywhere, but to complete the coverage, we can introduce some assertions on the sync targets passed to the callback. Up to you

syncing/src/test/kotlin/maru/syncing/beaconchain/CLSyncServiceImplTest.kt
```diff
@@ -0,0 +1,493 @@
...
+    for (i in 1uL..50uL) {
+      assertThat(beaconChain1.getSealedBeaconBlock(i)).isEqualTo(beaconChain2.getSealedBeaconBlock(i))
+      assertThat(beaconChain1.getBeaconState(i)).isEqualTo(beaconChain2.getBeaconState(i))
+    }
+
+    // update the sync target to 50 again
+    synced.set(false)
+    reset(beaconChain1, beaconChain2)
+    clSyncService.setSyncTarget(50uL)
+    awaitUntilAsserted { assertThat(synced).isTrue() }
```
Filter94, line 244: Should we await at all? If it's supposed to return immediately?

syncing/src/test/kotlin/maru/syncing/beaconchain/CLSyncServiceImplTest.kt
```diff
@@ -0,0 +1,493 @@
...
+      assertThat(beaconChain1.getBeaconState(i)).isEqualTo(beaconChain2.getBeaconState(i))
+    }
+
+    // Verify that beaconChain for node1 was not updated, and there are no reads on node2 beaconChain
+    verify(beaconChain1, never()).newUpdater()
+    verify(beaconChain2, never()).getSealedBeaconBlocks(any(), any())
+  }
+
+  @Test
+  fun `chain sync continues to download after stopped is called`() {
```
Filter94, line 294: I don't get why this is valid. I think it shouldn't download anything

syncing/src/test/kotlin/maru/syncing/beaconchain/CLSyncServiceImplTest.kt
```diff
@@ -0,0 +1,493 @@
...
+    assertThat(beaconChain1.getLatestBeaconState().latestBeaconBlockHeader.number).isEqualTo(100uL)
+    assertThat(beaconChain1.getLatestBeaconState()).isEqualTo(beaconChain2.getLatestBeaconState())
+    for (i in 1uL..100uL) {
+      assertThat(beaconChain1.getSealedBeaconBlock(i)).isEqualTo(beaconChain2.getSealedBeaconBlock(i))
+      assertThat(beaconChain1.getBeaconState(i)).isEqualTo(beaconChain2.getBeaconState(i))
+    }
+  }
+
+  @Test
+  fun `onSyncComplete handler is called only once per sync`() {
```
Filter94, line 324: I suggest adding:
1. A call to the same target
2. A call to the past target

syncing/src/test/kotlin/maru/syncing/beaconchain/CLSyncServiceImplTest.kt
```diff
@@ -0,0 +1,493 @@
...
+      assertThat(beaconChain1.getBeaconState(i)).isEqualTo(beaconChain2.getBeaconState(i))
+    }
+
+    // Verify that beaconChain for node1 was not updated, and there are no reads on node2 beaconChain
+    verify(beaconChain1, never()).newUpdater()
+    verify(beaconChain2, never()).getSealedBeaconBlocks(any(), any())
+  }
+
+  @Test
+  fun `sync target set to older target returns immediately`() {
```
Filter94, line 260: Can we please add the following test, similar to this one:
1. The source beacon chain only has blocks until 100
2. The target is set to 150
3. The target is set to 100
The expectation that the target beacon chain is synced up to a block 100

?

syncing/src/test/kotlin/maru/syncing/beaconchain/CLSyncServiceImplTest.kt
Filter94, line 68: Very nice testsuite overall!

Filter94 (approved):
I'm happy to merge it as is and address all the feedback in one of the follow-up PRs

consensus/src/main/kotlin/maru/consensus/qbft/DelayedQbftBlockCreator.kt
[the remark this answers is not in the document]
jframe: Not in the same module so internal can't be used in this case

syncing/src/main/kotlin/maru/syncing/beaconchain/CLSyncServiceImpl.kt
```diff
@@ -0,0 +1,105 @@
...
+      name = "beaconchain.restart.counter",
+      description = "Count of chain pipeline restarts",
+    )
+
+  override fun setSyncTarget(syncTarget: ULong) {
+    log.info("Syncing started syncTarget={}", syncTarget)
+    this.syncTarget.set(syncTarget)
+
+    // If the pipeline is already running, we don't need to start a new one
+    if (pipeline == null) {
```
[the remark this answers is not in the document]
jframe: Don't need `pipeline.abort` the `BeaconChainDownloadPipelineFactory` handles this in the iterator and will return immediately. The handler will still be called in this case.

syncing/src/test/kotlin/maru/syncing/beaconchain/CLSyncServiceImplTest.kt
```diff
@@ -0,0 +1,493 @@
...
+    validators = setOf(Validator(Util.publicKeyToAddress(keypair.publicKey).toArray()))
+
+    val genesisTimestamp = DataGenerators.randomTimestamp()
+    val (genesisBeaconState, genesisBeaconBlock) = genesisState(genesisTimestamp, validators)
+    beaconChain1 = spy(InMemoryBeaconChain(genesisBeaconState, genesisBeaconBlock))
+    beaconChain2 = spy(InMemoryBeaconChain(genesisBeaconState, genesisBeaconBlock))
+
+    port1 = findFreePort()
+    port2 = findFreePort()
+    p2pNetwork1 = createNetwork(beaconChain1, key1, port1)
```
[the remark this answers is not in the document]
jframe: I tried that it didn't work well. I used the port property from `P2PNetworkImpl` to get the actual port used and it would work for first test and fail with null errors for other tests. Need a bit more work to make `P2PNetworkImpl.port` more robust before I use that approach.

syncing/src/test/kotlin/maru/syncing/beaconchain/CLSyncServiceImplTest.kt
jframe: There was a comment that @fluentcrafter left saying it shouldn't matter if we call `start()` before setting the sync target to trigger the sync. I think it responsible considering this a service that we need to call `start()` before it can be used to sync and in the stopped state it won't trigger a sync. Sounds like that's also what you expect. So going to change to back to that behaviour.

Issue #244, opened 2025-07-29
Periodically update peer status


jframe:
## Description
Implement periodic peer status updates in MaruPeerManager to maintain an accurate view of peer status

## Motivation
Currently, peer status is not updated after the initial connection. We need up to date peer status for each peer to be able to select peers that have the requested sync target.

## Tasks
- [ ] Design periodic update mechanism with configurable interval
- [ ] Implement peer status polling logic in MaruPeerManager
- [ ] Add timeout handling for unresponsive peers

## Acceptance criteria
- [ ] Peer status is updated at regular configurable intervals
- [ ] Unresponsive peers are detected and marked as inactive within reasonable time
- [ ] Unit tests cover status update logic and edge cases

## Risks
- [ ] Frequent status checks could create network overhead
- [ ] Concurrent status updates may require careful synchronization

## Remember to
 - [ ] Add the `documentation` label in case there is an impact on the documentation
 - [ ] Add `priority` and `team` labels
 - [ ] Add Task for updating the Runbook or adding/updating existing metrics and alerts.

PR #256, opened 2025-07-30
Status update is updated every 5 minutes

pinges:
#244 
This PR adds a regular update of the Status of peers. If the update does not happen within time the peer is disconnected.

### DIFF
--- a/p2p/src/main/kotlin/maru/p2p/MaruPeer.kt
+++ b/p2p/src/main/kotlin/maru/p2p/MaruPeer.kt
@@ -78,6 +78,10 @@ class DefaultMaruPeer(
     ),
   private val p2pConfig: P2P,
 ) : MaruPeer {
+  init {
+    delegatePeer.subscribeDisconnect { _, _ -> scheduler.shutdown() }
+  }
+
   private val log: Logger = LogManager.getLogger(this.javaClass)
   private val status = AtomicReference<Status?>(null)
   private var scheduledDisconnect: Optional<ScheduledFuture<*>> = Optional.empty()
--- a/p2p/src/main/kotlin/maru/p2p/MaruPeer.kt
+++ b/p2p/src/main/kotlin/maru/p2p/MaruPeer.kt
@@ -38,7 +38,7 @@ import tech.pegasys.teku.networking.p2p.rpc.RpcStreamController
 interface MaruPeer : Peer {
   fun getStatus(): Status?
 
-  fun sendStatus(): SafeFuture<Status>
+  fun sendStatus(): SafeFuture<Unit>
 
   fun updateStatus(newStatus: Status)
 
@@ -84,23 +84,25 @@ class DefaultMaruPeer(
 
   override fun getStatus(): Status? = status.get()
 
-  override fun sendStatus(): SafeFuture<Status> {
+  override fun sendStatus(): SafeFuture<Unit> {
     try {
       val statusMessage = statusMessageFactory.createStatusMessage()
       val sendRpcMessage: SafeFuture<Message<Status, RpcMessageType>> =
         sendRpcMessage(statusMessage, rpcMethods.status())
       scheduleDisconnectIfStatusNotReceived(p2pConfig.statusUpdate.timeout)
-      return sendRpcMessage.thenApply { message -> message.payload }.whenComplete { status, error ->
-        if (error != null) {
-          disconnectImmediately(Optional.of(DisconnectReason.REMOTE_FAULT), false)
-          if (error.cause !is PeerDisconnectedException) {
-            log.debug("Failed to send status message to peer={}: errorMessage={}", this.id, error.message, error)
+      return sendRpcMessage
+        .thenApply { message -> message.payload }
+        .whenComplete { status, error ->
+          if (error != null) {
+            disconnectImmediately(Optional.of(DisconnectReason.REMOTE_FAULT), false)
+            if (error.cause !is PeerDisconnectedException) {
+              log.debug("Failed to send status message to peer={}: errorMessage={}", this.id, error.message, error)
+            }
+          } else {
+            updateStatus(status)
+            scheduler.schedule(this::sendStatus, p2pConfig.statusUpdate.renewal.inWholeSeconds, TimeUnit.SECONDS)
           }
-        } else {
-          updateStatus(status)
-          scheduler.schedule(this::sendStatus, p2pConfig.statusUpdate.renewal.inWholeSeconds, TimeUnit.SECONDS)
-        }
-      }
+        }.thenApply {}
     } catch (e: Exception) {
       if (e.cause !is PeerDisconnectedException) {
         log.error("Failed to send status message to peer={}", id, e)
--- a/p2p/src/main/kotlin/maru/p2p/MaruPeer.kt
+++ b/p2p/src/main/kotlin/maru/p2p/MaruPeer.kt
@@ -9,7 +9,12 @@
 package maru.p2p
 
 import java.util.Optional
+import java.util.concurrent.Executors
+import java.util.concurrent.ScheduledExecutorService
+import java.util.concurrent.ScheduledFuture
+import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicReference
+import kotlin.stackTraceToString
 import maru.p2p.messages.BeaconBlocksByRangeRequest
 import maru.p2p.messages.BeaconBlocksByRangeResponse
 import maru.p2p.messages.Status
@@ -28,6 +33,9 @@ import tech.pegasys.teku.networking.p2p.rpc.RpcRequestHandler
 import tech.pegasys.teku.networking.p2p.rpc.RpcResponseHandler
 import tech.pegasys.teku.networking.p2p.rpc.RpcStreamController
 
+private const val STATUS_RENEWAL_SECONDS = 300L // 5 minutes
+const val STATUS_TIMEOUT_SECONDS = 10L
+
 interface MaruPeer : Peer {
   fun getStatus(): Status?
 
@@ -39,6 +47,8 @@ interface MaruPeer : Peer {
     startBlockNumber: ULong,
     count: ULong,
   ): SafeFuture<BeaconBlocksByRangeResponse>
+
+  fun scheduleDisconnectIfStatusNotReceived(delay: Long)
 }
 
 interface MaruPeerFactory {
@@ -61,9 +71,12 @@ class DefaultMaruPeer(
   private val delegatePeer: Peer,
   private val rpcMethods: RpcMethods,
   private val statusMessageFactory: StatusMessageFactory,
+  private val scheduler: ScheduledExecutorService = Executors.newSingleThreadScheduledExecutor(),
 ) : MaruPeer {
   private val log: Logger = LogManager.getLogger(this::javaClass)
+
   private val status = AtomicReference<Status?>(null)
+  private var scheduledDisconnect: Optional<ScheduledFuture<*>> = Optional.empty()
 
   override fun getStatus(): Status? = status.get()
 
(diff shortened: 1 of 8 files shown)


p2p/src/main/kotlin/maru/p2p/MaruPeer.kt
```diff
@@ -28,6 +33,9 @@ import tech.pegasys.teku.networking.p2p.rpc.RpcRequestHandler
 import tech.pegasys.teku.networking.p2p.rpc.RpcResponseHandler
 import tech.pegasys.teku.networking.p2p.rpc.RpcStreamController
 
+private const val STATUS_RENEWAL_SECONDS = 300L // 5 minutes
```
Filter94, line 36: We will need this configurable. For the testing we need much faster feedback loop

p2p/src/main/kotlin/maru/p2p/MaruPeer.kt
```diff
@@ -72,15 +85,37 @@ class DefaultMaruPeer(
...
       return SafeFuture.failedFuture(e)
     }
   }
 
   override fun updateStatus(status: Status) {
+    scheduledDisconnect.ifPresent { it.cancel(false) }
     this.status.set(status)
+    log.debug("Received status update from peer {}: {}", delegatePeer.id, status)
+    if (this.connectionInitiatedRemotely()) {
+      scheduleDisconnectIfStatusNotReceived(STATUS_RENEWAL_SECONDS + 10L)
```
Filter94, line 110: This 10L should also be named. And ideally, be configurable

p2p/src/main/kotlin/maru/p2p/MaruPeer.kt
```diff
@@ -72,15 +85,37 @@ class DefaultMaruPeer(
       val statusMessage = statusMessageFactory.createStatusMessage()
       val sendRpcMessage: SafeFuture<Message<Status, RpcMessageType>> =
         sendRpcMessage(statusMessage, rpcMethods.status())
-      return sendRpcMessage.thenApply { message -> message.payload }
+      scheduleDisconnectIfStatusNotReceived(STATUS_TIMEOUT_SECONDS)
+      return sendRpcMessage.thenApply { message -> message.payload }.whenComplete { status, error ->
+        if (error != null) {
+          this.disconnectImmediately(Optional.of(DisconnectReason.REMOTE_FAULT), false)
+          log.debug("Failed to send status message to peer {}: {}", this.id, error.stackTraceToString())
```
fluentcrafter, line 92: as mentioned before, it's a bad practice to call `stackTraceToString()`, it should be up to log4j to that regarding log4j.xml configs. 

better option:
```kotlin
 log.debug("Failed to send status message to peer={}: errorMessage= {}", this.id, error.message, error)
```

p2p/src/main/kotlin/maru/p2p/MaruPeer.kt
```diff
@@ -72,15 +85,37 @@ class DefaultMaruPeer(
       val statusMessage = statusMessageFactory.createStatusMessage()
       val sendRpcMessage: SafeFuture<Message<Status, RpcMessageType>> =
         sendRpcMessage(statusMessage, rpcMethods.status())
-      return sendRpcMessage.thenApply { message -> message.payload }
+      scheduleDisconnectIfStatusNotReceived(STATUS_TIMEOUT_SECONDS)
+      return sendRpcMessage.thenApply { message -> message.payload }.whenComplete { status, error ->
```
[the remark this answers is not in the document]
fluentcrafter: I also prefer that approach. Poller vs Push. Furthermore, allows each node to set the update frequency better. 

The current approach disconnects automatically if not status received, it's serves as "clean up" process which is good. In poller approach maybe we can doo the same if peer does not reply successfully in the last S seconds?  

Anyway, if we decide to go with pool approach I ok merging this for now.

p2p/src/main/kotlin/maru/p2p/MaruPeer.kt
```diff
@@ -72,15 +85,37 @@ class DefaultMaruPeer(
...
     } catch (e: Exception) {
       log.error("Failed to send status message to peer ${delegatePeer.id}", e)
       return SafeFuture.failedFuture(e)
     }
   }
 
   override fun updateStatus(status: Status) {
+    scheduledDisconnect.ifPresent { it.cancel(false) }
     this.status.set(status)
+    log.debug("Received status update from peer {}: {}", delegatePeer.id, status)
```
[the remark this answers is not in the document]
pinges: Done

p2p/src/main/kotlin/maru/p2p/MaruPeerManager.kt
```diff
@@ -115,23 +112,13 @@ class MaruPeerManager(
...
-          Optional.of(DisconnectReason.REMOTE_FAULT),
-          false,
-        )
-      }
-    }, STATUS_TIMEOUT_SECONDS, TimeUnit.SECONDS)
-  }
-
   override fun onDisconnect(peer: Peer) {
     connectedPeers.remove(peer.id)
+    log.info("Peer {} disconnected", peer.id)
```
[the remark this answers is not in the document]
pinges: Done

p2p/src/main/kotlin/maru/p2p/MaruPeer.kt
```diff
@@ -39,6 +47,8 @@ interface MaruPeer : Peer {
     startBlockNumber: ULong,
     count: ULong,
   ): SafeFuture<BeaconBlocksByRangeResponse>
+
+  fun scheduleDisconnectIfStatusNotReceived(delay: Long)
```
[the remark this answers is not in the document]
pinges: Done

p2p/src/main/kotlin/maru/p2p/MaruPeer.kt
pinges: Done

p2p/src/main/kotlin/maru/p2p/MaruPeer.kt
```diff
@@ -72,15 +85,37 @@ class DefaultMaruPeer(
...
+      }
     } catch (e: Exception) {
       log.error("Failed to send status message to peer ${delegatePeer.id}", e)
       return SafeFuture.failedFuture(e)
     }
   }
 
   override fun updateStatus(status: Status) {
+    scheduledDisconnect.ifPresent { it.cancel(false) }
     this.status.set(status)
```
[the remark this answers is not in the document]
pinges: done

p2p/src/main/kotlin/maru/p2p/MaruPeer.kt
```diff
@@ -48,22 +56,30 @@ interface MaruPeerFactory {
...
   private val statusMessageFactory: StatusMessageFactory,
+  private val scheduler: ScheduledExecutorService =
+    Executors.newSingleThreadScheduledExecutor(
+      Thread.ofVirtual().factory(),
+    ),
+  private val p2pConfig: P2P,
 ) : MaruPeer {
   private val log: Logger = LogManager.getLogger(this.javaClass)
   private val status = AtomicReference<Status?>(null)
+  private var scheduledDisconnect: Optional<ScheduledFuture<*>> = Optional.empty()
```
Filter94, line 82: There can only be 1 scheduled disconnect at a time?

p2p/src/main/kotlin/maru/p2p/MaruPeer.kt
```diff
@@ -72,15 +88,41 @@ class DefaultMaruPeer(
       val statusMessage = statusMessageFactory.createStatusMessage()
       val sendRpcMessage: SafeFuture<Message<Status, RpcMessageType>> =
         sendRpcMessage(statusMessage, rpcMethods.status())
```
Filter94, line 90: I think we should handle failures here, during `sendRpcMessage` call. Because if we have a network disconnect, we'll fail to send the message. This call will fail and the next status update call won't ever be scheduled

p2p/src/main/kotlin/maru/p2p/MaruPeer.kt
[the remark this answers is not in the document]
jframe: I prefer to current approach of scheduling the updates individually. By individually doing the updates will avoid sending a large number of status requests at once. And we will get incremental updates of the peer status more often than one central status update loop.

p2p/src/main/kotlin/maru/p2p/MaruPeer.kt
```diff
@@ -72,15 +88,41 @@ class DefaultMaruPeer(
...
+      scheduleDisconnectIfStatusNotReceived(p2pConfig.statusUpdateTimeout)
+      return sendRpcMessage.thenApply { message -> message.payload }.whenComplete { status, error ->
+        if (error != null) {
+          disconnectImmediately(Optional.of(DisconnectReason.REMOTE_FAULT), false)
+          log.debug("Failed to send status message to peer={}: errorMessage={}", this.id, error.message, error)
+        } else {
+          updateStatus(status)
+          scheduler.schedule(this::sendStatus, p2pConfig.statusUpdateRenewal.inWholeSeconds, TimeUnit.SECONDS)
+        }
+      }
```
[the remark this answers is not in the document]
jframe: these issues look valid

p2p/src/main/kotlin/maru/p2p/MaruPeer.kt
```diff
@@ -72,15 +88,41 @@ class DefaultMaruPeer(
...
+    if (connectionInitiatedRemotely()) {
+      scheduleDisconnectIfStatusNotReceived(p2pConfig.statusUpdateRenewal + p2pConfig.statusUpdateLeeway)
+    }
+  }
+
+  override fun scheduleDisconnectIfStatusNotReceived(delay: Duration) {
+    scheduledDisconnect.ifPresent { it.cancel(false) }
+    scheduledDisconnect =
+      Optional.of(
+        scheduler.schedule(
```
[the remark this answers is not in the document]
pinges: I'm not sure I understand the question. If we are not disconnected when the next status update is triggered, we have to send the update, because otherwise the other peer is going to disconnect us. The other peer is expecting a status update within 5 minutes + leeway.
If the peer has been disconnected, it does not matter, because we can't send an update.

p2p/src/main/kotlin/maru/p2p/MaruPeer.kt
pinges: If we have a network disconnect we WANT to disconnect the peer, as it is useless.

Filter94: Let me clarify. If there's a network disruption for 100 milliseconds, all the peers will be lost immediately? There's 0 grace period?

pinges: We could retry to send the message, but I'm not sure it is worth it. We do have many peers and we will find another peer to fill the gap.

p2p/src/main/kotlin/maru/p2p/MaruPeer.kt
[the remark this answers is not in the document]
Filter94: 1. Should we call `scheduler.shutdown()` after `disconnectCleanly`?
2. Should we prevent the following `sendStatus` calls when `disconnectCleanly` has been called?

p2p/src/test/kotlin/maru/p2p/P2PTest.kt
```diff
@@ -894,12 +889,23 @@ class P2PTest {
...
-      while (System.currentTimeMillis() < startTime + 100000L) {
+      while (System.currentTimeMillis() < startTime + 11000L) {
         assertNetworkIsConnectedToPeer(p2pNetworkImpl1, PEER_ID_NODE_2)
         assertNetworkIsConnectedToPeer(p2pNetworkImpl2, PEER_ID_NODE_1)
+        peer1List.add(peer1.getStatus()!!.latestBlockNumber)
+        peer2List.add(peer2.getStatus()!!.latestBlockNumber)
+        sleep(500)
       }
+
+      assertThat(peer1List).contains(0uL, 1uL, 2uL, 3uL, 4uL, 5uL, 6uL, 7uL, 8uL, 9uL)
```
Filter94, line 907: It would be nice to make this logic more cohesive by reusing the exact same list that is used for mocking and tying its size to the `11000L` above that is used in a loop end condition

Filter94 (approved):

PR #260, opened 2025-07-31
Fixing flacky test

Filter94:
Failed test [run](https://github.com/Consensys/maru/actions/runs/16655684252/job/47139973173) 
Error:
```
SyncControllerThreadSafetyTest > should maintain EL-follows-CL invariant under concurrent access() FAILED
    org.opentest4j.AssertionFailedError: 
    expected: 0
     but was: 1
        at app//maru.syncing.SyncControllerThreadSafetyTest.should maintain EL-follows-CL invariant under concurrent access(SyncControllerThreadSafetyTest.kt:152)
```

I think it could happen because the reads to el and cl states are separate and not atomic

### DIFF
--- a/syncing/src/test/kotlin/maru/syncing/SyncControllerThreadSafetyTest.kt
+++ b/syncing/src/test/kotlin/maru/syncing/SyncControllerThreadSafetyTest.kt
@@ -15,7 +15,6 @@ import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicInteger
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.BeforeEach
-import org.junit.jupiter.api.RepeatedTest
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.parallel.Execution
 import org.junit.jupiter.api.parallel.ExecutionMode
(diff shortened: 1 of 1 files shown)


syncing/src/test/kotlin/maru/syncing/SyncControllerThreadSafetyTest.kt
```diff
@@ -115,8 +115,9 @@ class SyncControllerThreadSafetyTest {
     // Monitor for invariant violations: when CL starts syncing, EL should never be SYNCED
     executor.submit {
       repeat(iterations * 3) {
-        val clStatus = syncController.getCLSyncStatus()
-        val elStatus = syncController.getElSyncStatus()
+        val stateSnapshot = syncController.captureStateSnapshot()
```
gauravahuja, line 118: Does it make sense to remove getCLSyncStatus and getELSyncStatus methods from the interface. The interface can only expose getSyncState that captures both atomically.

Filter94: Definitely not as a part of this PR. But in general if we were to make changes here, I'd change the state into a 1 enum of 3 states: `SYNCING`, `CL_SYNCED`, `EL_AND_CL_SYNCED`

gauravahuja (approved):

PR #262, opened 2025-08-01
Integration of the syncing mechanism with Maru

Filter94:
Implements #228
* Introduced `awaitTillMaruHasPeers(1u)` awaits where Maru has to be synced at the start
* Introduced `PeeringNetworkParticipantStack` which is needed in cases where Besu nodes have to be started as the same cluster in order to peer and sync during the test
* Introduced `finalizationProvider` to `ElSyncService` implementation as it has to respect the finalization status
* Moved BeaconChain initialization to `MaruAppFactory` simplifying QBFT protocols
* Replaced a few `*::class.java` with `this.javaClass` in some places where `Logging` is instantiated to fix the logging issues
* Added `AlwaysSyncedController` for the tests without any P2P logic
* Removed tracking of the `currentSyncTarget` from `BeaconSyncControllerImpl` and `ELSyncService`
* Added `syncing.el-sync-status-refresh-interval` config
* Removed `QbftConsensusFollower` moved p2p topic registration to `QbftFollowerFactory` on MaruStartup instead
* Added a few more logs to `P2PNetworkImpl`
* `PeerChainTracker` now just sets the sync target if there are no peers
* Moved `PeerChainTrackerTest.TestableTimer` out from the class to the package for reuse


syncing/src/main/kotlin/maru/syncing/PeerChainTracker.kt
```diff
@@ -69,14 +69,17 @@ class PeerChainTracker(
...
+    val newSyncTarget =
+      if (peers.isNotEmpty()) {
+        targetChainHeadCalculator.selectBestSyncTarget(peers.values.toList())
+      } else {
+        0UL // If there are no peers, let's just say we're synced, because we don't know better
       }
+    log.trace("Selected best syncTarget={} lastNotifiedTarget={}", newSyncTarget, lastNotifiedTarget)
+    if (newSyncTarget != lastNotifiedTarget) { // Only send an update if there's an actual target change
+      beaconSyncTargetUpdateHandler.onBeaconChainSyncTargetUpdated(newSyncTarget)
+      log.trace("Notified about the new syncTarget={}", newSyncTarget)
```
fluentcrafter, line 81: maybe should be in info/debug mode? It's logged only when the target is updated, so it should not be that noisy and I think it would be usefull

app/src/main/kotlin/maru/app/MaruAppFactory.kt
```diff
@@ -300,4 +311,22 @@ class MaruAppFactory {
   private fun ensureDirectoryExists(path: Path) {
     if (!path.exists()) Files.createDirectories(path)
   }
+
+  private fun dbInitialization(
+    beaconGenesisConfig: ForksSchedule,
+    beaconChain: BeaconChain,
+  ) {
+    val qbftForkConfig = beaconGenesisConfig.getForkByConfigType(QbftConsensusConfig::class)
```
jframe, line 319: With this change we will assume that we always have Qbft in the config? Seems like a reasonable assumption. I think we should validating the config somewhere early to ensure that QBFT is configured. Not important for this PR though.

app/src/main/kotlin/maru/app/QbftProtocolFactoryWithBeaconChainInitialization.kt
```diff
@@ -63,8 +62,6 @@ class QbftProtocolFactoryWithBeaconChainInitialization(
         executionLayerEngineApiClient = engineApiExecutionLayerClient,
       )
 
-    beaconChainInitialization.ensureDbIsInitialized(validatorSet = qbftConsensusConfig.validatorSet)
```
jframe, line 66: Can shorten the name of this class now to `QbftProtocolFactory`

syncing/src/main/kotlin/maru/syncing/ELSyncService.kt
```diff
@@ -61,41 +63,37 @@ class ELSyncService(
...
-        blockNumber = latestSealedBeaconBlock.beaconBlock.beaconBlockBody.executionPayload.blockNumber,
-        blockHash = latestSealedBeaconBlock.beaconBlock.beaconBlockBody.executionPayload.blockHash,
+        blockNumber = latestBeaconBlockBody.executionPayload.blockNumber,
+        blockHash = latestBeaconBlockBody.executionPayload.blockHash,
       )
-    if (lastELSyncTarget != newELSyncTarget) {
-      lastELSyncTarget = newELSyncTarget
-      log.info("New EL sync target = {}", newELSyncTarget)
-    }
+    log.info("New EL sync target = {}", newELSyncTarget)
```
jframe, line 88: How often will this log be printed? It might be better at debug level if it's a lot or use the log throttle

app/src/main/kotlin/maru/app/MaruAppFactory.kt
```diff
@@ -65,7 +67,7 @@ import tech.pegasys.teku.networking.p2p.network.config.GeneratingFilePrivateKeyS
 import org.hyperledger.besu.plugin.services.MetricsSystem as BesuMetricsSystem
 
 class MaruAppFactory {
```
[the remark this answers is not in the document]
Filter94: I like `this.javaClass` because it's more copy-pasteable, it's shorter and if doesn't change when you rename the class.

It doesn't transfer well to companion objects, that's true

syncing/src/main/kotlin/maru/syncing/PeerChainTracker.kt
Filter94: Ok, I'll make it debug

app/src/main/kotlin/maru/app/MaruApp.kt
```diff
@@ -160,6 +160,13 @@ class MaruApp(
     beaconChain.close()
   }
 
+  fun peersConnected(): UInt =
```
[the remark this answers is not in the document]
Filter94: You're right. I think we could have used the REST API here, but I won't do that in the same PR

fluentcrafter (approved):

PR #263, opened 2025-08-01
subscription manager improvement

fluentcrafter:
Add the following improvements:

- subscriberId is optional - uses reflection or stack tracing to get subscriber's code origin 
- adds `interface Observable<T>` - components can leverage Kotlin implementation by delegation and don't need to write custom e.g `onStatusUpde(handler: (Status) -> Unit)`
- adds support to remove all subscriptions by reference `observable.removeSubscription(this)`

## Usage Example - Observable

```kotlin
class MyObservableClass(
  val elSyncSbsManager: SubscriptionManager<ELStatus>,
  val clSycnSbsManager: SubscriptionManager<CLStatus>,
) : Observable<ELStatus> by elSyncSbsManager, Observable<clSycnSbsManager> {
   fun setNewSyncTarget() {
    // ... logic
     elSyncSbsManager.notifySubscribers(ELStatus.SYNCHING)
     clSyncSbsManager.notifySubscribers(CLStatus.SYNCHING)
   }
}
```

## Usage Example - Observer

```kotlin
class StatusObserver(
 val statusObservable: Observable<ELStatus>,
) {
  fun start() {
    statusObservable.addSubscriber(this::onELStatusUpdate)
  }

 fun stop() {
    statusObservable.removeSubscriber(this::onELStatusUpdate)
  }


  fun onELStatusUpdate(newElStatus: ELStatus) {
  // ... logic
  }
}
```


core/src/main/kotlin/maru/subscription/SubscriptionManager.kt
```diff
@@ -10,21 +10,102 @@ package maru.subscription
...
+  return runCatching { receiverField.get(subscriberMethodRef) }.getOrNull()
+}
+
+@OptIn(ExperimentalReflectionOnLambdas::class)
+private fun <T> getSubscriberId(subscriber: (T) -> Any?): String {
+  val subscriberId =
+    when {
+      subscriber is kotlin.reflect.KFunction<*> && subscriber is CallableReference -> {
+        // we need to make the reference unique, so will use receiver object hashcode
+        val receiverObj = findReceiverObjRef(subscriber)!!
```
jframe, line 35: Can you just use `CallableReference.getBoundReceiver` here instead doing getting the field using reflection?

core/src/main/kotlin/maru/subscription/SubscriptionManager.kt
```diff
@@ -10,21 +10,102 @@ package maru.subscription
...
+        "${subscriber.owner.toString().replace("class ", "")}.${subscriber.name}@${receiverObj.hashCode()}"
```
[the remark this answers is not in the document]
jframe: Don't know enough of Kotlin internals to know if this is a legitimate issue, but sounds plausible that there could be no receiver object as `CallableReference` looks like it has support for that case.

core/src/main/kotlin/maru/subscription/SubscriptionManager.kt
```diff
@@ -10,21 +10,102 @@ package maru.subscription
...
   )
 
+  /**
+   * Adds a synchronous subscriber to the subscription manager.
+   * The subscriber will be notified in a blocking manner when data is published.
+   * The subscriberId is used to uniquely identify the subscriber for logging purposes when an error occurs.
+   *
+   * A subscriberId is generated by reflection automatically.
+   * Note: caller cannot subscribe 2 lambdas in the same line of code
+   * because it will generate the same subscriberId.
```
jframe, line 95: If 2 lambdas are subscribed from the same line of code will an error occur?

fluentcrafter: Yes, it's a limitation. To overcome it would require some overengineering and the gain is not worthy the effort ...

core/src/main/kotlin/maru/subscription/SubscriptionManager.kt
fluentcrafter: Yes, that's a good one. Thanks

Filter94 (approved):

PR #274, opened 2025-08-05
E2e sync from scratch with maru

Filter94:
* Integrated Maru into "switch from scratch E2E tests"
* Introduced a name label to `FollowerBeaconBlockImporter` for better logging
* Some refactoring of `CliqueToPosTest`
* Moved `switchTimestamp` to `MaruFactory`'s constructor as an optional argument
* Added scheduler shutdown on `MaruPeerManager` stop
* Fixed a race condition in `CLSyncServiceImpl` when `setTarget` is called right after completion of the pipeline
* Added newPayload call to `ELSyncService`, otherwise some EL clients won't sync. Added some additional logging there
* Improved `Timer()` threads naming
* Turned all the worker threads I could find into daemon threads to avoid JVM hanging
* Closing as much resources on close as possible
* Improved log collection in `CliqueToPosTest`. Now logs are collected for each run


consensus/src/main/kotlin/maru/consensus/blockimport/BeaconBlockImporter.kt
```diff
@@ -59,7 +62,9 @@ class FollowerBeaconBlockImporter(
       .newPayload(executionPayload)
       .handleException { e ->
         log.error(
-          "Error importing execution payload for blockNumber=${executionPayload.blockNumber}",
+          "Error importing execution payload to {} for blockNumber={}",
```
fluentcrafter, line 65:
```suggestion
          "Error importing execution payload to {} for elBlockNumber={}",
```

consensus/src/main/kotlin/maru/consensus/blockimport/BeaconBlockImporter.kt
```diff
@@ -70,6 +75,7 @@ class FollowerBeaconBlockImporter(
             safeHash = finalizationState.safeBlockHash,
             finalizedHash = finalizationState.finalizedBlockHash,
           ).thenApply {
+            log.debug("Imported blockNumber={} to {}", executionPayload.blockNumber, importerName)
```
fluentcrafter, line 78:
```suggestion
            log.debug("Imported elBlockNumber={} to {}", executionPayload.blockNumber, importerName)
```

jvm-libs/test-utils/src/main/kotlin/testutils/maru/MaruFactory.kt
```diff
@@ -103,7 +127,7 @@ class MaruFactory(
       SyncingConfig(
         peerChainHeightPollingInterval = 1.seconds,
         peerChainHeightGranularity = 1u,
-        elSyncStatusRefreshInterval = 500.milliseconds,
+        elSyncStatusRefreshInterval = 3.seconds,
```
fluentcrafter, line 130: why increase the refresh interval?

Filter94: It's a leftover. I'll change it back

executionlayer/src/main/kotlin/maru/executionlayer/client/PragueWeb3JJsonRpcExecutionLayerEngineApiClient.kt
```diff
@@ -106,4 +106,8 @@ class PragueWeb3JJsonRpcExecutionLayerEngineApiClient(
       emptyList(),
       Bytes32.ZERO,
     )
+
+  override fun close() {
+    web3jClient.eth1Web3j.shutdown()
```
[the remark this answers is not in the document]
Filter94: Yes, that's true. I'll pass the web3j client itself, same as I did with the ethApi client to the App so it can close it

p2p/src/main/kotlin/maru/p2p/MaruPeerManager.kt
```diff
@@ -33,7 +33,8 @@ import tech.pegasys.teku.networking.p2p.peer.NodeId
 import tech.pegasys.teku.networking.p2p.peer.Peer
 
 class MaruPeerManager(
-  private val scheduler: ScheduledExecutorService = Executors.newSingleThreadScheduledExecutor(),
+  private val scheduler: ScheduledExecutorService =
+    Executors.newSingleThreadScheduledExecutor(Thread.ofPlatform().daemon().factory()),
```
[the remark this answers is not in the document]
Filter94: Yes, that's the main motivation. My understanding is that with the old approach this thread would prevent JVM from exiting. And it's terminating by itself if it's for longer than 10 seconds without a task to execute

syncing/src/main/kotlin/maru/syncing/ELSyncService.kt
```diff
@@ -85,21 +92,37 @@ class ELSyncService(
...
 
     val finalizationState = finalizationProvider(latestBeaconBlockBody)
     val fcuResponse =
       executionLayerManager
-        .setHead(
-          headHash = newELSyncTarget.blockHash,
-          safeHash = finalizationState.safeBlockHash,
-          finalizedHash = finalizationState.finalizedBlockHash,
-        ).get()
+        .newPayload(latestBeaconBlockBody.executionPayload)
```
jonesho, line 100: Is this `newPayload` call a must to sync the EL? If not, why add here? Its a bit counter-intuitive to me, because how could EL validate the `latestBeaconBlockBody.executionPayload` (i.e. EL block 100) beforehand if it was not synced to that point (i.e. still at EL block 80)? Or I'm misunderstanding sth?

jonesho: Oh I found [this](https://github.com/danielrachi1/diagrams/blob/main/rendered/engine-api/newPayloadV2.png), and seems like if the EL doesn't have the required block data to validate the new payload, it would automatically switch to SYNCING mode (for some clients)

PR #275, opened 2025-08-06
Update elsync onchange and peerchaintracker no peer case

gauravahuja:


syncing/src/main/kotlin/maru/syncing/ELSyncService.kt
```diff
@@ -70,8 +70,10 @@ class ELSyncService(
     val latestBeaconBlockNumber = latestBeaconBlockHeader.number
     if (latestBeaconBlockNumber == 0UL) {
       val newELSyncStatus = ELSyncStatus.SYNCED
-      currentELSyncStatus = newELSyncStatus
-      onStatusChange(newELSyncStatus)
+      if (currentELSyncStatus != newELSyncStatus) {
```
Filter94, line 73: Minor: maybe we can move it to a separate function?

Filter94: Also, I think this deduplication of updates is breaking the tests, because I think at the time of startup not all of the handlers are set

syncing/src/main/kotlin/maru/syncing/SyncController.kt
```diff
@@ -85,38 +85,31 @@ class BeaconSyncControllerImpl(
...
 
-        // When CL starts syncing, EL must also be syncing (EL follows CL rule)
-        val newElStatus =
-          when {
-            newStatus == CLSyncStatus.SYNCING -> ELSyncStatus.SYNCING
-            else -> previousState.elStatus
-          }
-
-        currentState = SyncState(newStatus, newElStatus)
+        currentState = SyncState(newStatus, previousState.elStatus)
```
Filter94, line 94:
```suggestion
        currentState = currentState.copy(clStatus = newStatus)
```

Filter94 (approved):

PR #278, opened 2025-08-07
Implement Shanghai Engine Api Client

gauravahuja:
Implements #289

### DIFF
--- a/executionlayer/src/main/kotlin/maru/executionlayer/client/ShanghaiWeb3JJsonRpcExecutionLayerEngineApiClient.kt
+++ b/executionlayer/src/main/kotlin/maru/executionlayer/client/ShanghaiWeb3JJsonRpcExecutionLayerEngineApiClient.kt
@@ -13,28 +13,28 @@ import maru.config.consensus.ElFork
 import maru.core.ExecutionPayload
 import maru.extensions.captureTimeSafeFuture
 import maru.mappers.Mappers.toDomainExecutionPayload
-import maru.mappers.Mappers.toExecutionPayloadV3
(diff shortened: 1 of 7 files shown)


executionlayer/src/main/kotlin/maru/executionlayer/client/ShanghaiWeb3JJsonRpcExecutionLayerEngineApiClient.kt
```diff
@@ -0,0 +1,81 @@
...
+import tech.pegasys.teku.ethereum.executionclient.schema.PayloadAttributesV1
+import tech.pegasys.teku.ethereum.executionclient.schema.PayloadAttributesV2
+import tech.pegasys.teku.ethereum.executionclient.schema.PayloadStatusV1
+import tech.pegasys.teku.ethereum.executionclient.schema.Response
+import tech.pegasys.teku.ethereum.executionclient.web3j.Web3JClient
+import tech.pegasys.teku.infrastructure.async.SafeFuture
+import tech.pegasys.teku.infrastructure.bytes.Bytes8
+
+// https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md
+class ShanghaiWeb3JJsonRpcExecutionLayerEngineApiClient(
```
Filter94, line 29: It would be nice to test it with a real node. For the `PragueWeb3JJsonRpcExecutionLayerEngineApiClient` it was done by `EagerQbftBlockCreatorTest`, because `EagerQbftBlockCreator` is supposed to implement the correct sequence of calls, which serves as a test for the client itself.

It's Ok to do it in a follow-up PR

Filter94 (approved):

PR #279, opened 2025-08-07
fix flaky discovery test

pinges:
I've corrected the test `discovery finds nodes` to check the right thing, added continuous searching, and fixed bootnode enr

### DIFF
--- a/p2p/src/main/kotlin/maru/p2p/MaruPeerManager.kt
+++ b/p2p/src/main/kotlin/maru/p2p/MaruPeerManager.kt
@@ -48,6 +48,7 @@ class MaruPeerManager(
   private val log: Logger = LogManager.getLogger(this.javaClass)
   private val maxPeers = p2pConfig.maxPeers
   private val currentlySearching = AtomicBoolean(false)
+  private val connectedPeers: ConcurrentHashMap<NodeId, MaruPeer> = ConcurrentHashMap()
 
   private val connectionInProgress = mutableListOf<Bytes>()
   private var discoveryService: MaruDiscoveryService? = null
@@ -100,8 +101,6 @@ class MaruPeerManager(
       }
     }
   }
-
-  private val connectedPeers: ConcurrentHashMap<NodeId, MaruPeer> = ConcurrentHashMap()
 
   private fun logConnectedPeers() {
     val peerIds = connectedPeers.keys.joinToString(", ") { it.toString() }
--- a/p2p/src/main/kotlin/maru/p2p/MaruPeerManager.kt
+++ b/p2p/src/main/kotlin/maru/p2p/MaruPeerManager.kt
@@ -142,10 +142,10 @@ class MaruPeerManager(
         return
       }
       synchronized(connectionInProgress) {
-        if (p2pNetwork.peerCount >= maxPeers || connectionInProgress.contains(peer.nodeIdBytes)) {
+        if (p2pNetwork.peerCount >= maxPeers || connectionInProgress.contains(peer.nodeId)) {
           return
         }
-        connectionInProgress.add(peer.nodeIdBytes)
+        connectionInProgress.add(peer.nodeId)
       }
 
       p2pNetwork
@@ -154,18 +154,18 @@ class MaruPeerManager(
         .whenComplete { _, throwable ->
           try {
             if (throwable != null) {
-              log.error("Failed to connect to peer={}", peer.nodeIdBytes, throwable)
+              log.error("Failed to connect to peer={}", peer.nodeId, throwable)
             }
           } finally {
             synchronized(connectionInProgress) {
-              connectionInProgress.remove(peer.nodeIdBytes)
+              connectionInProgress.remove(peer.nodeId)
             }
           }
         }
     } catch (e: Exception) {
-      log.debug("Failed to initiate connection to peer={}. errorMessage={}", peer.nodeIdBytes, e.message, e)
+      log.debug("Failed to initiate connection to peer={}. errorMessage={}", peer.nodeId, e.message, e)
       synchronized(connectionInProgress) {
-        connectionInProgress.remove(peer.nodeIdBytes)
+        connectionInProgress.remove(peer.nodeId)
       }
     }
   }
--- a/p2p/src/main/kotlin/maru/p2p/MaruPeerManager.kt
+++ b/p2p/src/main/kotlin/maru/p2p/MaruPeerManager.kt
@@ -68,6 +68,7 @@ class MaruPeerManager(
 
   fun stop(): SafeFuture<Unit> {
     stopCalled = true
+    scheduler.shutdown()
     return SafeFuture.completedFuture(Unit)
   }
 
--- a/p2p/src/main/kotlin/maru/p2p/MaruPeerManager.kt
+++ b/p2p/src/main/kotlin/maru/p2p/MaruPeerManager.kt
@@ -33,22 +33,18 @@ import tech.pegasys.teku.networking.p2p.peer.NodeId
 import tech.pegasys.teku.networking.p2p.peer.Peer
 
 class MaruPeerManager(
-  private val scheduler: ScheduledExecutorService =
-    Executors.newSingleThreadScheduledExecutor(Thread.ofPlatform().daemon().factory()),
   private val maruPeerFactory: MaruPeerFactory,
   private val p2pConfig: P2P,
 ) : PeerHandler,
   PeerLookup {
-  init {
-    scheduler.scheduleAtFixedRate({
-      logConnectedPeers()
-    }, 20000, 20000, TimeUnit.MILLISECONDS)
-  }
+  var scheduler: ScheduledExecutorService? = null
 
   private val log: Logger = LogManager.getLogger(this.javaClass)
   private val maxPeers = p2pConfig.maxPeers
   private val currentlySearching = AtomicBoolean(false)
   private val connectedPeers: ConcurrentHashMap<NodeId, MaruPeer> = ConcurrentHashMap()
+  val peerCount: Int
+    get() = connectedPeers.size
 
   private val connectionInProgress = mutableListOf<Bytes>()
   private var discoveryService: MaruDiscoveryService? = null
@@ -56,32 +52,46 @@ class MaruPeerManager(
   private lateinit var p2pNetwork: P2PNetwork<Peer>
 
   @Volatile
-  private var stopCalled = false
+  private var started = AtomicBoolean(false)
 
   fun start(
     discoveryService: MaruDiscoveryService?,
     p2pNetwork: P2PNetwork<Peer>,
   ) {
+    if (!started.compareAndSet(false, true)) {
+      log.warn("Trying to start already started MaruPeerManager")
+      return
+    }
     this.discoveryService = discoveryService
     this.p2pNetwork = p2pNetwork
+    if (scheduler == null) {
+      scheduler = Executors.newSingleThreadScheduledExecutor(Thread.ofPlatform().daemon().factory())
+    }
+    scheduler!!.scheduleAtFixedRate({
+      logConnectedPeers()
+    }, 20000, 20000, TimeUnit.MILLISECONDS)
     searchForPeersUntilMaxReached()
   }
 
   fun stop(): SafeFuture<Unit> {
-    stopCalled = true
-    scheduler.shutdown()
+    if (!started.compareAndSet(true, false)) {
+      log.warn("Trying to stop stopped MaruPeerManager")
+      return SafeFuture.completedFuture(Unit)
+    }
+    scheduler!!.shutdown()
+    scheduler = null
     return SafeFuture.completedFuture(Unit)
   }
 
   private fun searchForPeersUntilMaxReached() {
-    if (!stopCalled && currentlySearching.compareAndSet(false, true)) {
+    if (started.get() && currentlySearching.compareAndSet(false, true)) {
       discoveryService?.let { discoveryService ->
         discoveryService
           .searchForPeers()
           .orTimeout(Duration.ofSeconds(30L))
           .whenComplete { availablePeers, throwable ->
             log.trace("Finished searching for peers. Found {} peers.", availablePeers.size)
-            if (!stopCalled) {
+            if (started.get()) {
               if (throwable != null) {
                 log.debug("Failed to discover peers: {}", throwable.message, throwable)
                 discoveryService.getKnownPeers()
@@ -93,9 +103,9 @@ class MaruPeerManager(
             }
           }.whenComplete { _, _ ->
             currentlySearching.set(false)
-            log.trace("peerCount={}. maxPeers={}", p2pNetwork.peerCount, maxPeers)
-            if (!stopCalled && p2pNetwork.peerCount < maxPeers) {
-              scheduler.schedule({ searchForPeersUntilMaxReached() }, 1, TimeUnit.SECONDS)
+            log.trace("peerCount={}. maxPeers={}", peerCount, maxPeers)
+            if (started.get() && peerCount < maxPeers) {
+              scheduler!!.schedule({ searchForPeersUntilMaxReached() }, 1, TimeUnit.SECONDS)
             }
           }
       }
@@ -110,12 +120,12 @@ class MaruPeerManager(
 
   override fun onConnect(peer: Peer) {
     // TODO: here we could check if we want to be connected to that peer
-    if (p2pNetwork.peerCount > maxPeers) {
+    if (peerCount > maxPeers) {
       // TODO: We could disconnect another peer here, based on some criteria
       peer.disconnectCleanly(DisconnectReason.TOO_MANY_PEERS)
     }
     val maruPeer = maruPeerFactory.createMaruPeer(peer)
-    connectedPeers.put(peer.id, maruPeer)
+    connectedPeers[peer.id] = maruPeer
     if (maruPeer.connectionInitiatedLocally()) {
       maruPeer.sendStatus()
     } else {
@@ -126,7 +136,7 @@ class MaruPeerManager(
   override fun onDisconnect(peer: Peer) {
     connectedPeers.remove(peer.id)
     log.trace("Peer={} disconnected", peer.id)
-    if (!stopCalled && p2pNetwork.peerCount < maxPeers) {
+    if (started.get() && peerCount < maxPeers) {
       searchForPeersUntilMaxReached()
     }
   }
@@ -138,11 +148,11 @@ class MaruPeerManager(
   private fun tryToConnectIfNotFull(peer: MaruDiscoveryPeer) {
     try {
       val peerId = getNodeId(peer)
-      if (stopCalled || p2pNetwork.isConnected(PeerAddress(peerId))) {
+      if (!started.get() || p2pNetwork.isConnected(PeerAddress(peerId))) {
         return
       }
       synchronized(connectionInProgress) {
-        if (p2pNetwork.peerCount >= maxPeers || connectionInProgress.contains(peer.nodeId)) {
+        if (peerCount >= maxPeers || connectionInProgress.contains(peer.nodeId)) {
           return
         }
         connectionInProgress.add(peer.nodeId)
--- a/p2p/src/main/kotlin/maru/p2p/P2PNetworkImpl.kt
+++ b/p2p/src/main/kotlin/maru/p2p/P2PNetworkImpl.kt
@@ -161,7 +161,7 @@ class P2PNetworkImpl(
       }
 
   override fun stop(): SafeFuture<Unit> {
-    log.trace("Stopping {}", this::class.simpleName)
+    log.info("Stopping {}", this::class.simpleName)
     val pmStop = maruPeerManager.stop()
     discoveryService?.stop()
     val p2pStop = p2pNetwork.stop()
@@ -284,7 +284,7 @@ class P2PNetworkImpl(
         .toUInt()
 
   internal val peerCount: Int
-    get() = p2pNetwork.peerCount
+    get() = maruPeerManager.getPeers().size
 
   internal fun isConnected(peer: String): Boolean {
     val peerAddress =
--- a/p2p/src/main/kotlin/maru/p2p/P2PNetworkImpl.kt
+++ b/p2p/src/main/kotlin/maru/p2p/P2PNetworkImpl.kt
@@ -114,7 +114,6 @@ class P2PNetworkImpl(
         privateKeyBytes = privateKeyBytesWithoutPrefix(privateKeyBytes),
         p2pConfig = p2pConfig,
         forkIdHashProvider = forkIdHashProvider,
-        metricsSystem = metricsSystem,
       )
     }
 
@@ -131,7 +130,7 @@ class P2PNetworkImpl(
 
   override val nodeId: String = p2pNetwork.nodeId.toBase58()
   override val discoveryAddresses: List<String> = p2pNetwork.discoveryAddresses.getOrElse { emptyList() }
-  override val enr: String? = p2pNetwork.enr.getOrNull()
+  override val enr: String? = discoveryService?.getLocalNodeRecord()?.asEnr()
   override val nodeAddresses: List<String> = p2pNetwork.nodeAddresses
 
   override fun start(): SafeFuture<Unit> =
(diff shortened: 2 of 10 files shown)


p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
```diff
@@ -59,15 +60,15 @@ class MaruDiscoveryService(
       .map { NodeRecordFactory.DEFAULT.fromEnr(it) }
       .toList()
 
-  private val discoverySystem: DiscoverySystem =
+  val discoverySystem: MutableDiscoverySystem =
```
Filter94, line 63: If you need `nodeRecordBuckets` from `discoverySystem` can you please expose it via `internal` method rather than making `discoverySystem` public?

p2p/src/test/kotlin/maru/p2p/EnrStringUtil.kt
```diff
@@ -25,7 +25,7 @@ fun getBootnodeEnrString(
   val bootnodeNR =
     NodeRecordBuilder()
       .nodeRecordFactory(NodeRecordFactory(IdentitySchemaInterpreter.V4))
-      .seq(1)
```
Filter94, line 28: Can you please explain this change?

p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
```diff
@@ -45,7 +46,7 @@ class MaruDiscoveryService(
   }
 
   companion object {
-    const val FORK_ID_HASH_FIELD_NAME = "mfidh"
+    const val FORK_ID_HASH_FIELD_NAME = "eth2"
```
Filter94, line 49: Why are we changing it to `eth2`?

p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
```diff
@@ -80,22 +81,14 @@ class MaruDiscoveryService(
...
-    discoveryNetworkBuilder.secretKey(privateKey)
-    discoveryNetworkBuilder.localNodeRecord(localNodeRecord())
-    discoveryNetworkBuilder.bootnodes(bootnodes)
-  }
-
   fun start() {
     discoverySystem
       .start()
       .thenRun {
         this.bootnodeRefreshTask =
```
Filter94, line 88: Can you please check the pattern we use in `PeerChainTracker` and `ELSyncService`? We inject `timerFactory` there and there's a testable implementation `TestablePeriodicTimer`, which rules out any inconsistency in the tests

p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
Filter94: Or even better you can also write the logic to collect all nodes from different buckets into the same collection so it's easier to check in the test

p2p/src/test/kotlin/maru/p2p/discovery/MaruDiscoveryServiceTest.kt
```diff
@@ -248,32 +246,55 @@ class MaruDiscoveryServiceTest {
       discoveryService2.start()
       discoveryService3.start()
 
+      continuouslySearchForPeersWithDelay(bootnode)
+      continuouslySearchForPeersWithDelay(discoveryService2)
+      continuouslySearchForPeersWithDelay(discoveryService3)
```
Filter94, line 251: Since we are in controlled environment and we know that all 3 services are up, why do we want to _continuously_ search for peers. And why do we use `discoveryService`'s internal executor for that? It breaks encapsulation

p2p/src/test/kotlin/maru/p2p/P2PTest.kt
```diff
@@ -1157,8 +1157,8 @@ class P2PTest {
     peer: String,
   ) {
     assertThat(
-      p2pNetwork.isConnected(peer),
-    ).isTrue()
+      p2pNetwork.getPeerLookup().getPeer(LibP2PNodeId(PeerId.fromBase58(peer)))
```
Filter94, line 1160: Now it doesn't verify what the name of the function claims

p2p/src/test/kotlin/maru/p2p/discovery/MaruDiscoveryServiceTest.kt
```diff
@@ -248,32 +246,55 @@ class MaruDiscoveryServiceTest {
...
+        var found = false
+        nodeRecordBuckets.forEach { bucket ->
+          bucket.forEach { node ->
+            if (node.nodeId == expectedNodeId) {
+              log.info("Found peer: ${node.nodeId}")
+              found = true
+            }
+          }
+        }
+        assertThat(found).isTrue()
```
Filter94, line 297: Is it easier to use `find()` instead?

p2p/src/test/kotlin/maru/p2p/EnrStringUtil.kt
pinges: in discovery v5 there is a mechanism to update the node. The ping message contains the sequence number. If the seq number is higher than the one in the buckets the latest version is requested.

p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
pinges: I was thinking about this. Now I'm not sure about this anymore. eth2 is the field that they are using in the CLs. I thought that when we are using eth2 we could find out when there are a lot of CLs trying to connect. I think I will change that back ...

p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
pinges: That was just for debugging ... Will change it back!

p2p/src/test/kotlin/maru/p2p/EnrStringUtil.kt
Filter94: What is a node update?

p2p/src/test/kotlin/maru/p2p/discovery/MaruDiscoveryServiceTest.kt
pinges: This is just for testing, but you are right, I shouldn't use the internal executor. 
The search that is happening here is the way to actively find new peers. In Maru this is done by the MaruPeerManager, when we need more peers.

Filter94: I get it, but if you want to test that specifically, it's logical to do it in `MaruPeerManager`'s tests. IMO, in `MaruDiscoveryServiceTest` you should focus on its public api, which is `searchForPeers`

p2p/src/test/kotlin/maru/p2p/P2PTest.kt
Filter94: You mean that if a peer is found in the lookup, it means it's connected?

p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
pinges: done

p2p/src/test/kotlin/maru/p2p/discovery/MaruDiscoveryServiceTest.kt
pinges: Hmm, you are right ... There is only 3 nodes here and in theory it should be enough to call search for peers once, as the bootnode should already be discovered. I'll change the test.

p2p/src/test/kotlin/maru/p2p/P2PTest.kt
pinges: Yes. Peer lookup uses the connectedPeers from the MaruPeerManager, which contains the peers that we are connected to.

p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
```diff
@@ -63,53 +66,33 @@ class MaruDiscoveryService(
...
+            /* period = */ p2pConfig.discovery!!.refreshInterval.inWholeMilliseconds,
           )
-      }.get(DISCOVERY_TIMEOUT_SECONDS, TimeUnit.SECONDS)
+      }
     return
   }
 
-  fun stop() {
-    bootnodeRefreshTask.cancel()
+  override fun stop() {
```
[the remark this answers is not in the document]
pinges: canceling the poller now

p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
```diff
@@ -63,53 +66,33 @@
...
```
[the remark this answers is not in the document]
pinges: cancelling the poller now

p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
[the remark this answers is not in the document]
pinges: fixed

p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
```diff
@@ -63,54 +65,32 @@ class MaruDiscoveryService(
...
+        )
+      }
     return
   }
 
-  fun stop() {
-    bootnodeRefreshTask.cancel()
+  override fun stop() {
     discoverySystem.stop()
+    poller?.cancel()
```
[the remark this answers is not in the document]
pinges: the poller scheduled ping is not a necessity to start the discovery service.
I'll change the stop to cancel the poller first.

p2p/src/test/kotlin/maru/p2p/discovery/MaruDiscoveryServiceTest.kt
```diff
@@ -236,47 +216,39 @@ class MaruDiscoveryServiceTest {
...
+        .searchForPeers()
+        .thenAccept { foundPeers ->
+          assertThat(foundPeers.any { it.nodeId == bootnode.getLocalNodeRecord().nodeId }).isTrue
+          assertThat(foundPeers.any { it.nodeId == discoveryService3.getLocalNodeRecord().nodeId }).isTrue
+        }.join()
+      discoveryService3
+        .searchForPeers()
+        .thenAccept { foundPeers ->
+          assertThat(foundPeers.any { it.nodeId == bootnode.getLocalNodeRecord().nodeId }).isTrue
+          assertThat(foundPeers.any { it.nodeId == discoveryService2.getLocalNodeRecord().nodeId }).isTrue
```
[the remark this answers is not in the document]
pinges: was the same thing, but cleaned that up.

p2p/src/main/kotlin/maru/p2p/P2PNetworkImpl.kt
```diff
@@ -131,7 +130,7 @@ class P2PNetworkImpl(
 
   override val nodeId: String = p2pNetwork.nodeId.toBase58()
   override val discoveryAddresses: List<String> = p2pNetwork.discoveryAddresses.getOrElse { emptyList() }
-  override val enr: String? = p2pNetwork.enr.getOrNull()
+  override val enr: String? = discoveryService?.getLocalNodeRecord()?.asEnr()
```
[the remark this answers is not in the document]
pinges: p2pNetwork.enr.getOrNull() always returned null, so this is better.

p2p/src/test/kotlin/maru/p2p/EnrStringUtil.kt
jframe: We don't have ping message. Should the current seq number be added to the status message since we are using that like a ping message?

p2p/src/test/kotlin/maru/p2p/discovery/MaruDiscoveryServiceTest.kt
```diff
@@ -236,47 +216,39 @@ class MaruDiscoveryServiceTest {
...
     try {
       bootnode.start()
       discoveryService2.start()
       discoveryService3.start()
 
-      awaitPeerFound(bootnode, discoveryService2.getLocalNodeRecord().nodeId)
-      awaitPeerFound(discoveryService2, discoveryService3.getLocalNodeRecord().nodeId)
-      awaitPeerFound(discoveryService3, discoveryService2.getLocalNodeRecord().nodeId)
-      awaitPeerFound(bootnode, discoveryService3.getLocalNodeRecord().nodeId)
+      // make sure services are started and bootnodes have been pinged
```
Filter94, line 230: But with the injectable timerFactory you can force it without the sleep. That's the whole point of making it injectable

p2p/src/main/kotlin/maru/p2p/MaruPeerManager.kt
```diff
@@ -142,10 +142,10 @@ class MaruPeerManager(
         return
       }
       synchronized(connectionInProgress) {
-        if (p2pNetwork.peerCount >= maxPeers || connectionInProgress.contains(peer.nodeIdBytes)) {
+        if (p2pNetwork.peerCount >= maxPeers || connectionInProgress.contains(peer.nodeId)) {
```
Filter94, line 145:
```suggestion
        if (getPeers().size >= maxPeers || connectionInProgress.contains(peer.nodeId)) {
```
Since `p2pNetwork` implements it via `MaruPeerManager`, using `p2pNetwork` here doesn't seem right

p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
```diff
@@ -198,11 +185,13 @@ class MaruDiscoveryService(
     )
   }
 
-  private fun localNodeRecord(): NodeRecord {
+  private fun createLocalNodeRecord(): NodeRecord {
     val nodeRecordBuilder: NodeRecordBuilder =
       NodeRecordBuilder()
         .secretKey(privateKey)
         .seq(UInt64.ONE)
```
Filter94, line 192: Does it have to be 1 or 0?

p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
```diff
@@ -63,53 +66,37 @@ class MaruDiscoveryService(
...
-                "Failed to contact discovery bootnodes",
-                error,
-              )
-            },
-          )
-      }.get(DISCOVERY_TIMEOUT_SECONDS, TimeUnit.SECONDS)
+        poller = timerFactory("boot-node-refresher", true)
+
+        poller!!.scheduleAtFixedRate(
+          /* task = */ timerTask { pingBootnodes() },
```
Filter94, line 90: We better wrap it in a try catch, at least to print uncaught exceptions

p2p/src/main/kotlin/maru/p2p/MaruPeerManager.kt
```diff
@@ -68,6 +69,7 @@ class MaruPeerManager(
 
   fun stop(): SafeFuture<Unit> {
     stopCalled = true
+    scheduler.shutdown()
```
Filter94, line 72: I didn't add it here intentionally. We rely on the following rule: `start() can be called after stop()`, but `scheduler` can't be started back up again after `shutdown` call. So either we create the `scheduler` inside `start`, or we don't terminate it on `stop`

p2p/src/test/kotlin/maru/p2p/discovery/MaruDiscoveryServiceTest.kt
```diff
@@ -135,10 +117,10 @@ class MaruDiscoveryServiceTest {
           P2P.Discovery(
             port = 9000u,
             bootnodes = listOf(),
-            refreshInterval = 5.seconds,
+            refreshInterval = 10.seconds,
           ),
       )
-    service = MaruDiscoveryService(keyPair.secretKey().bytesArray(), p2pConfig, forkIdHashProvider, NoOpMetricsSystem())
+    service = MaruDiscoveryService(keyPair.secretKey().bytesArray(), p2pConfig, forkIdHashProvider)
```
Filter94, line 123: Minor: Can we please add named args here

p2p/src/test/kotlin/maru/p2p/discovery/MaruDiscoveryServiceTest.kt
```diff
@@ -236,47 +216,39 @@ class MaruDiscoveryServiceTest {
...
-      awaitPeerFound(bootnode, discoveryService3.getLocalNodeRecord().nodeId)
+      // make sure services are started and bootnodes have been pinged
+      sleep(3000)
+
+      discoveryService2
+        .searchForPeers()
+        .thenAccept { foundPeers ->
+          assertThat(foundPeers.any { it.nodeId == bootnode.getLocalNodeRecord().nodeId }).isTrue
+          assertThat(foundPeers.any { it.nodeId == discoveryService3.getLocalNodeRecord().nodeId }).isTrue
+        }.join()
```
Filter94, line 238: join() will keep this test running forever if it's not true. Please use get() with a timeout

p2p/src/test/kotlin/maru/p2p/discovery/MaruDiscoveryServiceTest.kt
```diff
@@ -236,47 +216,39 @@ class MaruDiscoveryServiceTest {
...
+      discoveryService3
+        .searchForPeers()
+        .thenAccept { foundPeers ->
+          assertThat(foundPeers.any { it.nodeId == bootnode.getLocalNodeRecord().nodeId }).isTrue
```
Filter94, line 242: Minor: You can factor these assertions or the `searchForPeers` call with assertions to a function to make the tests a bit more readable

p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
```diff
@@ -63,53 +66,37 @@ class MaruDiscoveryService(
...
-      "DiscoveryService",
-      1,
-      1,
-      5,
-      MetricTrackingExecutorFactory(metricsSystem),
-    )
+  private var poller: Timer? = null
 
-  private lateinit var bootnodeRefreshTask: Cancellable
+  private val isStarted = AtomicBoolean(false)
```
Filter94, line 75: Why do you need this if you have `poller` instance? `poller` is supposed to be `null` when it's stopped, if you manage concurrency in `start()` and `stop()` properly? Also, `isStarted` doesn't seem to be used anywhere and it's not reset in `stop()`, as cursor bot noted

p2p/src/test/kotlin/maru/p2p/discovery/MaruDiscoveryServiceTest.kt
```diff
@@ -192,11 +174,10 @@ class MaruDiscoveryServiceTest {
               P2P.Discovery(
                 port = PORT2,
                 bootnodes = emptyList(),
-                refreshInterval = 5.seconds,
+                refreshInterval = 10.seconds,
```
fluentcrafter, line 177: why increasing the refresh interval? Tests whould have fast timers and run as fast as possible, otherwise build times keep growing

p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
```diff
@@ -24,20 +28,20 @@ import org.ethereum.beacon.discovery.schema.EnrField
...
   private val forkIdHashProvider: ForkIdHashProvider,
-  metricsSystem: MetricsSystem,
-) {
+  private val timerFactory: (String, Boolean) -> Timer = { name, isDaemon ->
+    Timer(
+      "$name-${UUID.randomUUID()}",
+      isDaemon,
+    )
+  },
+) : LongRunningService {
```
fluentcrafter, line 44: suggestion: create a factory function into a companion object and make it the default. It will improve readability


```kotlin
class MaruDiscoveryService(
  privateKeyBytes: ByteArray,
  private val p2pConfig: P2P,
  private val forkIdHashProvider: ForkIdHashProvider,
  private val createTimmerFn: (String, Boolean) -> Timer = createTimmer
) : LongRunningService {

  companion object {
    fun createTimmer(namePrefix: String, isDaemon: Boolean): Timer {
      return Timer(
      "$name-${UUID.randomUUID()}",
      isDaemon,
    )
    }
}
```

p2p/src/main/kotlin/maru/p2p/MaruPeerManager.kt
pinges: fixed

p2p/src/main/kotlin/maru/p2p/MaruPeerManager.kt
pinges: Done

p2p/src/test/kotlin/maru/p2p/discovery/MaruDiscoveryServiceTest.kt
pinges: done

p2p/src/test/kotlin/maru/p2p/discovery/MaruDiscoveryServiceTest.kt
pinges: It is just waiting for the thenAccept completable future, which completes after the two assertsThat

p2p/src/test/kotlin/maru/p2p/discovery/MaruDiscoveryServiceTest.kt
pinges: done

p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
pinges: Done

p2p/src/test/kotlin/maru/p2p/discovery/MaruDiscoveryServiceTest.kt
pinges: This is not a timer. The test will finish once the conditions are asserted, it will not wait for 10 seconds. Setting the refresh to 1 second doesn't make the test finish earlier. The refresh timer is used to ping the bootnodes every x seconds, to get our node into the buckets, so the bootnode can tell other nodes about our existence. The initial ping happens without delay for the first call, and then every x seconds. In this test only the initial ping is needed.

p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
pinges: good idea

pinges (commented):
@fluentcrafter 
A bootnode needs to have discovery enabled. It does discovery in the same way as all other nodes, the only difference is that it is known to the other nodes so these other nodes can contact it as a starting point of finding other peers. 
The IP address of the bootnode is configured in the p2p config.

p2p/src/test/kotlin/maru/p2p/discovery/MaruDiscoveryServiceTest.kt
Filter94: But how do you know if `searchForPeers` ever ends?

p2p/src/test/kotlin/maru/p2p/discovery/MaruDiscoveryServiceTest.kt
Filter94: I get it now. `pingBootnodes` doesn't actually complete the whole ping. it just sends out the requests and sets up the response handlers. This issue would be solvable if we used a dedicated interface tailored for our business object instead of plain `NodeRecord` and had a way to inject them into `MaruDiscoveryService`

p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
```diff
@@ -183,9 +187,9 @@ class MaruDiscoveryService(
     return true
   }
 
-  private fun pingBootnodes(bootnodeRecords: List<NodeRecord>) {
+  private fun pingBootnodes() {
     log.trace("Pinging bootnodes")
-    bootnodeRecords.forEach(
+    bootnodes.forEach(
       Consumer { bootnode: NodeRecord? ->
```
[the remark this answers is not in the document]
pinges: Done

PR #284, opened 2025-08-08
Maru follower syncing integration test

gauravahuja:


app/src/integrationTest/kotlin/maru/app/MaruFollowerTest.kt
```diff
@@ -247,4 +323,18 @@ class MaruFollowerTest {
...
       }
   }
+
+  private fun checkStackBlocks(
+    stack: PeeringNodeNetworkStack,
+    blocksToProduce: Int,
+  ) {
+    await
+      .pollDelay(100.milliseconds.toJavaDuration())
+      // we need big timeout due to CI resources sometimes being slow
```
Filter94, line 333: This comment might be irrelevant anymore. You can try to remove it

app/src/integrationTest/kotlin/maru/app/MaruFollowerTest.kt
```diff
@@ -247,4 +323,18 @@ class MaruFollowerTest {
         assertThat(blocksImportedByFollower).isEqualTo(blocksProducedByQbftValidator)
       }
   }
+
+  private fun checkStackBlocks(
```
Filter94, line 327: I didn't catch from the name that you meant "network stack". Can you please add it to the name? Or better, keep the name simple `checkBlocksOn` and rename `stack` to `networkStack` and `blocksToProduce` to `blocksExpected`

app/src/integrationTest/kotlin/maru/app/MaruFollowerTest.kt
```diff
@@ -236,6 +236,82 @@ class MaruFollowerTest {
...
+      }
+    }
+
+    // This is here mainly to wait until block propagation is complete
+    checkStackBlocks(validatorStack, blocksToProduce)
+
+    followerStack.maruApp.stop()
+    followerStack.maruApp.close()
+    followerStack.setMaruApp(
+      maruFactory.buildTestMaruFollowerWithP2pPeering(
```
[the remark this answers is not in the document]
gauravahuja: done

Filter94:
In all of the `MaruFollowerTest` when Maru instances are restarted, `peerChainHeightGranularity` is left as default, which gives false positives. And with `peerChainHeightGranularity` introduced, we should either be very deliberate about the height of the chain at the time of the sync and/or we should produce more blocks after restart to see that the node doesn't miss blocks after the restart.

app/src/integrationTest/kotlin/maru/app/MaruFollowerTest.kt
```diff
@@ -236,15 +237,140 @@ class MaruFollowerTest {
...
+      ),
+    )
+    followerStack.maruApp.start()
+
+    checkValidatorAndFollowerBlocks(blocksToProduce)
+  }
+
+  @Test
+  fun `Maru follower is able to complete syncing after restarted`() {
+    val blocksToProduce = 20
```
jonesho, line 273: Imo, setting `blocksToProduce = 10` should be enough for this test, as we will do two rounds

app/src/integrationTest/kotlin/maru/app/MaruFollowerTest.kt
```diff
@@ -236,15 +237,140 @@ class MaruFollowerTest {
...
+    blocksProduced: Int,
+  ) {
+    await
+      .pollDelay(100.milliseconds.toJavaDuration())
+      .timeout(30.seconds.toJavaDuration())
+      .untilAsserted {
+        val blocksOnStack = stack.besuNode.getMinedBlocks(blocksProduced)
+        assertThat(blocksOnStack.size).isEqualTo(blocksProduced)
+      }
+  }
```
jonesho, line 375: nit: just a small comment, I would recommend to extend this to 
```
private fun checkNetworkStackBlocksProduced(
    stacks: List<PeeringNodeNetworkStack>,
    blocksProduced: Int,
  ) 
```
so we could remove the use of `checkValidatorAndFollowerBlocks` and replace with `checkNetworkStackBlocksProduced`

jonesho (approved):
LGTM, minor comments left

Issue #155, opened 2025-06-20
Rpc message sending compression


jframe:
Currently P2P RPC and Gossip P2P messages are exchanged without any compression between the nodes. We should enable snappy compression for smaller network footprint. I think it has to be implemented as a generic Serializer and Deserialized that is delegates the actual serialization to a delegate and compressed whatever it gets from the delegate. You can refer to Teku or Besu code to see how they're doing it for inspiration

PR #291, opened 2025-08-08
feat: added compression on gossip and P2P response messages

jonesho:
This implements https://github.com/Consensys/maru/issues/155


p2p/src/main/kotlin/maru/p2p/messages/StatusMessageSerDe.kt
```diff
@@ -8,16 +8,42 @@
...
+import maru.serialization.rlp.RLPSerDe
+import org.hyperledger.besu.ethereum.rlp.RLPInput
+import org.hyperledger.besu.ethereum.rlp.RLPOutput
 
 class StatusMessageSerDe(
-  private val statusSerDe: SerDe<Status>,
-) : SerDe<Message<Status, RpcMessageType>> {
-  override fun serialize(value: Message<Status, RpcMessageType>): ByteArray = statusSerDe.serialize(value.payload)
+  private val statusSerDe: RLPSerDe<Status>,
+  compressor: MaruCompressor = MaruNoOpCompressor(),
```
jonesho, line 23: Here we use `MaruNoOpCompressor` because the compressed size for Status message is going to be larger than the raw size (i.e. few bytes larger), we might even consider just to extend `RLPSerDe` instead of `MaruCompressorRLPSerDe`

serialization/src/main/kotlin/maru/serialization/rlp/CompressorRLPSerDe.kt
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright Consensys Software Inc.
+ *
+ * This file is dual-licensed under either the MIT license or Apache License 2.0.
+ * See the LICENSE-MIT and LICENSE-APACHE files in the repository root for details.
+ *
+ * SPDX-License-Identifier: MIT OR Apache-2.0
+ */
+package maru.serialization.rlp
+
+interface CompressorRLPSerDe<T> : RLPSerDe<T> {
```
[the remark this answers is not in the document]
Filter94: Actually, you already have the `MaruCompressorRLPSerDe`, which could be just a class if you apply my suggestion

serialization/src/test/kotlin/maru/compression/MaruSnappyFramedCompressorTest.kt
```diff
@@ -0,0 +1,44 @@
...
+  @Test
+  fun `can compress and decompress zero size payload`() {
+    val payload = ByteArray(0)
+    val compressedPayload = compressor.compress(payload)
+    val decompressedPayload = compressor.decompress(compressedPayload)
+    Assertions.assertThat(decompressedPayload).isEqualTo(payload)
+  }
+
+  @Test
+  fun `throw exception if compress and decompress oversized payload`() {
```
Filter94, line 38: I would add a valid big payload case as well

serialization/src/main/kotlin/maru/serialization/rlp/CompressorRLPSerDe.kt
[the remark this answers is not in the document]
jonesho: Yes, I think your suggestion indeed makes sense, which we could remove this `CompressorRLPSerDe` entirely given that there's not much value on exposing this interface for classes to implement the `compress` and `decompress` functions

serialization/src/main/kotlin/maru/serialization/rlp/SealedBeaconBlockSerDe.kt
```diff
@@ -8,14 +8,17 @@
...
 import maru.core.SealedBeaconBlock
+import maru.serialization.compression.MaruSnappyFramedCompressor
 import org.hyperledger.besu.ethereum.rlp.RLPInput
 import org.hyperledger.besu.ethereum.rlp.RLPOutput
 
 class SealedBeaconBlockSerDe(
   private val beaconBlockSerializer: BeaconBlockSerDe,
   private val sealSerializer: SealSerDe,
-) : RLPSerDe<SealedBeaconBlock> {
+  private val compressor: MaruCompressor = MaruSnappyFramedCompressor(),
```
jframe, line 20: Will this affect RLP going into the database? Generally better to let rocksdb do it's own compression.

jframe: If we have rocksdb compression on, we may end up double compressing this and increasing the size for blocks in the database

jframe: Checked we are using the default KVStoreConfiguration that has LZ4 compression enabled https://github.com/Consensys/teku/blob/master/storage/src/main/java/tech/pegasys/teku/storage/server/kvstore/KvStoreConfiguration.java#L105 in the KvDatabaseFactory https://github.com/Consensys/maru/blob/main/storage/src/main/kotlin/maru/database/kv/KvDatabaseFactory.kt#L28

p2p/src/main/kotlin/maru/p2p/messages/BeaconBlocksByRangeResponseMessageSerDe.kt
```diff
@@ -8,44 +8,40 @@
...
-        writeTo(value.payload, rlpOutput)
-      }.toArray()
-
-  override fun deserialize(bytes: ByteArray): Message<BeaconBlocksByRangeResponse, RpcMessageType> =
-    Message(RpcMessageType.BEACON_BLOCKS_BY_RANGE, Version.V1, readFrom(RLP.input(Bytes.wrap(bytes))))
-
-  private fun writeTo(
-    value: BeaconBlocksByRangeResponse,
+  compressor: MaruCompressor = MaruSnappyCompressor(),
+) : MaruCompressorRLPSerDe<Message<BeaconBlocksByRangeResponse, RpcMessageType>>(compressor) {
```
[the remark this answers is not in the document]
Filter94: Yes. Current approach works with complete `ByteArrays`, which can consume a lot of memory. If we could compress the `RpcStream` directly, it would be much nicer. But I don't really see how it should work for the gossiping, because `gossip` method accepts whole bytes

p2p/src/test/kotlin/maru/p2p/messages/StatusMessageSerDeTest.kt
```diff
@@ -0,0 +1,36 @@
...
+class StatusMessageSerDeTest {
+  private val serDe = StatusMessageSerDe(StatusSerDe())
+
+  @Test
+  fun `can serialize and deserialize same value`() {
+    val testValue =
+      Message(
+        RpcMessageType.STATUS,
+        Version.V1,
+        Status(Random.nextBytes(32), Random.nextBytes(32), Random.nextULong()),
```
Filter94, line 28: You can move  `DataGenerators.randomStatus()` to `maru.core.ext` and reuse it here

p2p/src/main/kotlin/maru/p2p/messages/BeaconBlocksByRangeResponseMessageSerDe.kt
```diff
@@ -8,44 +8,40 @@
...
-import maru.serialization.SerDe
+import maru.serialization.compression.MaruSnappyFramedCompressor
+import maru.serialization.rlp.MaruCompressorRLPSerDe
 import maru.serialization.rlp.SealedBeaconBlockSerDe
-import org.apache.tuweni.bytes.Bytes
-import org.hyperledger.besu.ethereum.rlp.RLP
 import org.hyperledger.besu.ethereum.rlp.RLPInput
 import org.hyperledger.besu.ethereum.rlp.RLPOutput
 
 class BeaconBlocksByRangeResponseMessageSerDe(
```
Filter94, line 21: This class doesn't have to implement the `MaruCompressorRLPSerDe`. You can just instantiate `MaruCompressorRLPSerDe` and pass it as an `RLPSerDe<BeaconBlocksByRangeResponse>`

serialization/src/main/kotlin/maru/serialization/rlp/SealedBeaconBlockSerDe.kt
jonesho: Yes, now I'd reverted the changes on `SealedBeaconBlockSerDe` (just extends `RLPSerDe` instead of `MaruCompressorRLPSerDe` to disable compression), so the "db double compression" issue should be resolved. And introduced `SealedBeaconBlockCompressorSerDe` for compression when transmitting the gossip message

Filter94 (dismissed):
LGTM

p2p/src/test/kotlin/maru/p2p/messages/BeaconBlocksByRangeResponseMessageSerDeTest.kt
```diff
@@ -12,15 +12,22 @@ import maru.core.ext.DataGenerators
...
 
 class BeaconBlocksByRangeResponseMessageSerDeTest {
-  private val messageSerDe = BeaconBlocksByRangeResponseMessageSerDe(RLPSerializers.SealedBeaconBlockSerializer)
-
   @Test
   fun `response message serDe serializes and deserializes correctly`() {
+    val messageSerDe =
+      BeaconBlocksByRangeResponseMessageSerDe(
+        beaconBlocksByRangeResponseSerDe =
+          MaruCompressorRLPSerDe(
```
jframe, line 26: This test shouldn't be using the compressor, we need know this works without the compressor as it's not built into the `BeaconBlocksByRangeResponseMessageSerDe`

jframe (dismissed):

Filter94 (approved):

PR #292, opened 2025-08-11
Update CliqueToPOSTest for Clique->Shanghai->Prague switch

gauravahuja:


core/src/main/kotlin/maru/consensus/ForkSpec.kt
```diff
@@ -26,27 +27,27 @@ class ForksSchedule(
...
 ) {
+  private val log = LogManager.getLogger(this.javaClass)
   private val forks: NavigableSet<ForkSpec> =
     run {
       val newForks =
         TreeSet(
-          Comparator.comparing(ForkSpec::timestampSeconds).reversed(),
+          Comparator
+            .comparing(ForkSpec::timestampSeconds)
+            .thenComparing { forks.indexOf(it) }
```
Filter94, line 37: Why is this needed? If we have 2 forks with the timestamp, it should be an error. We can even put a validation for it

core/src/main/kotlin/maru/consensus/ForkSpec.kt
```diff
@@ -26,27 +27,27 @@ class ForksSchedule(
...
+            .reversed(),
         )
       newForks.addAll(forks)
+      log.info("Initialized forks schedule={}", newForks)
```
Filter94, line 41: I suggest to rather implement proper `toString()` for the `ForksSchedule` and log it somewhere in the user code, for example in MaruApp, the same place where we log the configration

core/src/test/kotlin/maru/consensus/ForksScheduleTest.kt
```diff
@@ -21,14 +21,16 @@ class ForksScheduleTest {
   @Test
   fun `test getForkByTimestamp returns correct fork`() {
     val fork1 = ForkSpec(1L, 1, consensusConfig)
-    val fork2 = ForkSpec(2L, 2, consensusConfig)
+    val fork2A = ForkSpec(2L, 20, consensusConfig)
+    val fork2B = ForkSpec(2L, 21, consensusConfig)
```
Filter94, line 25: I don't think it's a valid case

docker/initialization/init.sh
```diff
@@ -12,6 +12,7 @@ if [[ ! -f "initialization/genesis-besu.json" && ! -f "initialization/genesis-ge
 
     merge_timestamp=$(($(date +%s) + 60))
     echo "Timestamp: $merge_timestamp"
+    sed -i "s/%SHANGHAI_TIME%/$merge_timestamp/g" genesis-maru.json
```
Filter94, line 15: EL genesis files have to be updated too to respect it

core/src/main/kotlin/maru/consensus/ForkSpec.kt
gauravahuja: We can't log it anywhere else as forks is private 
```
private val forks
```

Filter94: That's why I mentioned `toString()`

e2e/src/acceptanceTest/kotlin/maru/e2e/CliqueToPosTest.kt
```diff
@@ -411,7 +432,26 @@ class CliqueToPosTest {
...
+      ElFork.entries.associateWith {
+        val engineApiClient =
+          Helpers.buildExecutionEngineClient(
+            web3JEngineApiClient = web3jClient,
+            elFork = it,
+            metricsFacade = TestEnvironment.testMetricsFacade,
+          )
+        JsonRpcExecutionLayerManager(engineApiClient)
+      }
+    val forkScheduleAwareExecutionLayerManager =
```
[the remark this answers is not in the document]
gauravahuja: It will be used after the Shanghai Switch and then we can other other forks that require a different engine api version. Thus the name ForkScheduleAware.

core/src/main/kotlin/maru/consensus/ForkSpec.kt
Filter94, line 1: I suggest we rename this file to `ForkSchedule`, because it's more important than `ForkSpec`

executionlayer/src/main/kotlin/maru/executionlayer/manager/ForkScheduleAwareExecutionLayerManager.kt
```diff
@@ -0,0 +1,76 @@
...
+  val clock: Clock = Clock.System,
+) : ExecutionLayerManager {
+  init {
+    ElFork.entries.forEach {
+      require(executionLayerManagerMap.containsKey(it)) { "No execution layer manager provided for $it" }
+    }
+  }
+
+  internal fun getCurrentElFork(): ElFork {
+    val currentTimestamp = clock.now().epochSeconds
```
Filter94, line 30: I realized that this logic is somewhat working, but isn't good enough. One of the side effects of the ElSyncService is that it would start the El sync before the CL sync is complete. And if it sets the FCU to the past block heights, it needs to get the forks using the timestamps from the current state of the `BeaconChain`. This logic will start to work only when `BeaconChain` reaches the current fork

Filter94: > I do not understand the need to use the timestamp from the current state to decide on the EL Fork, the EL Node would switch independently based on the its clock and not the timestamp

Did you try it? Different newPayload methods need to be used depending on the EL node's current fork

executionlayer/src/main/kotlin/maru/executionlayer/manager/ForkScheduleAwareExecutionLayerManager.kt
```diff
@@ -0,0 +1,76 @@
...
+    val forkSpec = forksSchedule.getForkByTimestamp(currentTimestamp)
+    return when (val configuration = forkSpec.configuration) {
+      is QbftConsensusConfig -> configuration.elFork
+      else -> ElFork.entries.first()
```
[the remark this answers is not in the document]
gauravahuja: Yes, I am making that change

executionlayer/src/main/kotlin/maru/executionlayer/manager/ForkScheduleAwareExecutionLayerManager.kt
Filter94: But we want `SYNCING` response

Filter94 (approved):

PR #301, opened 2025-08-12
p2p discovery - fix by selecting interface ipaddess when config points 0.0.0.0

fluentcrafter:


p2p/src/test/kotlin/maru/p2p/NetworkHelperTest.kt
```diff
@@ -0,0 +1,85 @@
...
+import java.net.Inet4Address
+import java.net.InetAddress
+import org.assertj.core.api.Assertions.assertThat
+import org.assertj.core.api.Assertions.assertThatThrownBy
+import org.junit.jupiter.api.Test
+
+class NetworkHelperTest {
+  @Test
+  fun `listNetworkAddresses returns network addresses`() {
+    NetworkHelper.listNetworkAddresses(excludeLoopback = true).also { addresses ->
```
Filter94, line 20: This test passes on CI because the list is empty. It would be nice to run these tests on controlled data.

p2p/src/main/kotlin/maru/p2p/NetworkHelper.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    if (ips.contains(targetIpV4)) {
+      return targetIpV4
+    } else if (targetIpV4 == "0.0.0.0") {
+      return ips.first()
+    } else {
+      val ipsString = ips.joinToString(separator = ",") { it }
+      throw IllegalArgumentException("targetIpV4=$targetIpV4 not found in machine interfaces, available ips=$ipsString")
+    }
+  }
```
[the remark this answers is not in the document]
Filter94: If it's true, we better add a validation that p2p's host ip can't be a loopback address

p2p/src/test/kotlin/maru/p2p/P2PTest.kt
```diff
@@ -65,7 +65,7 @@ class P2PTest {
   companion object {
     private val chainId = 1337u
 
-    private const val IPV4 = "127.0.0.1"
+    private val IPV4: String = NetworkHelper.listIpsV4(excludeLoopback = true).first()
```
Filter94, line 68: How does it pass on CI of CI machines only have loopback addresses?

p2p/src/test/kotlin/maru/p2p/NetworkHelperTest.kt
Filter94: I'm not so sure now. Are there any non-loopback interfaces on CI?

> What do you have in mind for the controlled data?

The same list as in the k8s cluster, or even your local machine could be good enough

p2p/src/test/kotlin/maru/p2p/P2PTest.kt
fluentcrafter: > How does it pass on CI if CI machines only have loopback addresses?

where did you get that info that they one have loopback?

IFIK, GHA runners have access to the internet, so they need at least one other interface than loopback.

p2p/src/test/kotlin/maru/p2p/NetworkHelperTest.kt
fluentcrafter: Not worth the effort for no IMO, will leave it we have way more pressing matters

Filter94 (dismissed):

Filter94 (approved):

PR #302, opened 2025-08-12
MaruDiscoveryService - remove dirty reflection for tests

fluentcrafter:


jframe (approved):
That's much nicer removing that reflection hack

PR #305, opened 2025-08-13
chaos: allow image override and improv configs

fluentcrafter:
This adds:
- a Chaos testing workflow experiment (not complete)
- a `health-check` test to check if nodes are in sync after Chaos experiments are run
- make target to orchestrate deployments, kubectl port forwarding, and run the health check


.github/workflows/chaos-testing.yml
```diff
@@ -0,0 +1,65 @@
+name: Testing CI
+
+on:
+  workflow_dispatch:
+    inputs:
+      consensus-client-changed:
+        required: true
+        type: string
+        default: 'true'
+        description: 'when false, skips the job'
```
gauravahuja, line 10: Can you add some context on when it should be set to false? If it is supposed to be a simple run/skip the job why not name it accordingly.

chaos-testing/health-checker/src/main/kotlin/maru/clients/beacon/BeaconChainClient.kt
```diff
@@ -0,0 +1,149 @@
...
+import org.http4k.format.Jackson.auto
+import tech.pegasys.teku.infrastructure.async.SafeFuture
+
+data class Genesis(
+  val genesis_time: String,
+  val genesis_validators_root: String,
+  val genesis_fork_version: String,
+)
+
+data class SyncingStatus(
```
gauravahuja, line 27: Why do you need most of these data classes, they are already defined in the API package and can be reused. Also some data class attributes do not follow the camel case naming convention.

gauravahuja (approved):

.github/workflows/chaos-testing.yml
fluentcrafter: Yeah, it was a bit confusing to me as well. Also, this is a bit of an anti-pattern.

Although it follows what's already done on testing (it was C&) I and want to change/improve both... but first I need to puth this config into `main` to enable the GHA and them I can refactor and improve. 

So, this is just an initial definition which I don't event kno it will work. Will improve in a followup PR

PR #309, opened 2025-08-15
Import block to payload validator el node of a maru follower

gauravahuja:
- Import block to EL PayloadValidator node of a MaruFollower
- Move the creation of new block handlers to the ProtocolFactory as they have access to the forkSpec and create the correct Follower Engine API version.


app/src/main/kotlin/maru/app/QbftFollowerFactory.kt
```diff
@@ -43,8 +44,12 @@ class QbftFollowerFactory(
     val stateTransition = StateTransitionImpl(validatorProvider)
     val transactionalSealedBeaconBlockImporter =
       TransactionalSealedBeaconBlockImporter(beaconChain, stateTransition) { _, beaconBlock ->
-        newBlockHandler.handleNewBlock(beaconBlock) // Don't wait for the result
-        SafeFuture.completedFuture(Unit)
+        elPayloadValidatorNewBlockHandler
+          .handleNewBlock(beaconBlock)
+          .thenCompose {
```
Filter94, line 49: I think we don't really need thenCompose here

app/src/main/kotlin/maru/app/MaruApp.kt
```diff
@@ -248,10 +248,21 @@ class MaruApp(
           syncStatusProvider = syncStatusProvider,
         )
       } else {
+        val elPayloadValidatorHandler =
+          createElNodeHandlers(
+            mapOf("payload-validator" to config.validatorElNode.engineApiEndpoint),
+            beaconGenesisConfig,
+          )
+        val elPayloadValidatorNewBlockHandler =
+          NewBlockHandlerMultiplexer(elPayloadValidatorHandler + metadataCacheUpdaterHandlerEntry)
```
Filter94, line 257: `metadataCacheUpdaterHandlerEntry` needs to be updated after the validation, so it should be in the same map as `elFollowerHandlersMap` and I think this means that we can move the instantiation of the follower handlers out of both branches, same as it used to be

app/src/main/kotlin/maru/app/QbftFollowerFactory.kt
gauravahuja: Can you explain why? I thought we should only import the block to the el followers and update metadata after the payload validator has successfully imported it, thus `thenCompose`

app/src/main/kotlin/maru/app/QbftFollowerFactory.kt
```diff
@@ -8,70 +8,97 @@
...
+            callAndForgetNewBlockHandler.handleNewBlock(beaconBlock) // Don't wait for the result
+            SafeFuture.completedFuture(Unit)
+          }
```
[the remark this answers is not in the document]
Filter94: It should be Ok with the `NewBlockHandlerMultiplexer`, because all the handlers are wrapped in a try-catch inside

PR #312, opened 2025-08-15
feat: add configs to control TargetChainHeadCalculator and DownloadPe…

jonesho:
…erProvider

### DIFF
--- a/syncing/src/main/kotlin/maru/syncing/beaconchain/pipeline/DownloadBlocksStep.kt
+++ b/syncing/src/main/kotlin/maru/syncing/beaconchain/pipeline/DownloadBlocksStep.kt
@@ -23,8 +23,32 @@ import tech.pegasys.teku.networking.eth2.rpc.core.RpcException
 import tech.pegasys.teku.networking.p2p.peer.DisconnectReason
 import tech.pegasys.teku.networking.p2p.reputation.ReputationAdjustment
 
-class DownloadBlocksStep(
+interface DownloadPeerProvider {
+  fun getDownloadingPeer(targetBlockNumber: ULong): MaruPeer
+}
+
+class DownloadPeerProviderImpl(
   private val peerLookup: PeerLookup,
+  private val useUnconditionalRandomSelection: Boolean,
+) : DownloadPeerProvider {
+  override fun getDownloadingPeer(targetBlockNumber: ULong): MaruPeer =
+    peerLookup
+      .getPeers()
+      .let {
+        if (!useUnconditionalRandomSelection) {
+          it.filter { peer ->
+            peer.getStatus() != null &&
+              peer.getStatus()!!.latestBlockNumber >=
+              targetBlockNumber
+          }
+        } else {
+          it
+        }
+      }.random()
+}
+
+class DownloadBlocksStep(
+  private val downloadPeerProvider: DownloadPeerProvider,
   private val maxRetries: UInt,
   private val blockRangeRequestTimeout: Duration,
 ) : Function<SyncTargetRange, CompletableFuture<List<SealedBlockWithPeer>>> {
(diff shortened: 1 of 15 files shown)


syncing/src/test/kotlin/maru/syncing/beaconchain/pipeline/DownloadBlocksTest.kt
```diff
@@ -245,6 +245,70 @@ class DownloadBlocksTest {
     }
   }
 
+  @Test
+  fun `always select a random peer if there are available peers when purely random selection is true`() {
```
Filter94, line 249: I suggest to test it differently. Have a list of 1 peer, with `randomStatus(0U)` and the test would check if it actually returns the only available option. The filtering implementation would fail in this case

app/src/integrationTest/kotlin/maru/app/MaruFollowerTest.kt
Filter94, line 1: I don't think it's reasonable to test each configuration combination for each of the scenarios as an app test. I would just add 1 more test for the `(MostFrequent, true)` combination in the MaruFollowerTest or in a new testsuite if it makes it easier

Filter94:
I made a simplified way of configuring the multiple strategies. Can you pelase take a look? https://github.com/Consensys/maru/pull/316/files

syncing/src/main/kotlin/maru/syncing/beaconchain/pipeline/DownloadBlocksStep.kt
```diff
@@ -23,8 +23,32 @@ import tech.pegasys.teku.networking.eth2.rpc.core.RpcException
 import tech.pegasys.teku.networking.p2p.peer.DisconnectReason
 import tech.pegasys.teku.networking.p2p.reputation.ReputationAdjustment
 
-class DownloadBlocksStep(
+interface DownloadPeerProvider {
+  fun getDownloadingPeer(targetBlockNumber: ULong): MaruPeer
```
fluentcrafter, line 27:
```suggestion
  fun getDownloadingPeer(downloadRangeEndBlockNumber: ULong): MaruPeer
```

This helps to understand what his value is semantically. Initially, I thought this was the head sync target.

config/src/main/kotlin/maru/config/ConfigLoaderHelper.kt
```diff
@@ -34,7 +34,7 @@ fun ConfigLoaderBuilder.addTomlDecoders(strict: Boolean): ConfigLoaderBuilder =
 @OptIn(ExperimentalHoplite::class)
 inline fun <reified T : Any> parseConfig(
   toml: String,
-  strict: Boolean = true,
+  strict: Boolean = false,
```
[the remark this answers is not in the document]
jonesho: This is introduced by Roman's PR to simpify the config, Roman and I had the same discussion here: https://github.com/Consensys/maru/pull/316/files#r2286093798

app/src/integrationTest/kotlin/maru/app/MaruFollowerTest.kt
```diff
@@ -342,35 +400,48 @@ class MaruFollowerTest {
...
+  private fun checkNetworkStacksBlocksProduced(
     blocksProduced: Int,
+    vararg stacks: PeeringNodeNetworkStack,
   ) {
     await
       .pollDelay(100.milliseconds.toJavaDuration())
       .timeout(30.seconds.toJavaDuration())
       .untilAsserted {
-        val blocksOnStack = stack.besuNode.getMinedBlocks(blocksProduced)
-        assertThat(blocksOnStack.size).isEqualTo(blocksProduced)
```
jonesho, line 373: This was wrong because `stack.besuNode.getMinedBlocks(blocksProduced)` would always return a list of blocks with size equal to `blocksProduced`, and put null on those that were failed to get, therefore I made changes here: https://github.com/Consensys/maru/pull/312/files#:~:text=%7D.mapNotNull%20%7B%20it.get().block%20%7D

PR #340, opened 2025-08-25
Constant round timer expiry

Filter94:


config/src/main/kotlin/maru/config/Config.kt
```diff
@@ -71,7 +71,6 @@ data class ValidatorElNode(
 data class QbftOptions(
   val minBlockBuildTime: Duration = 500.milliseconds,
   val messageQueueLimit: Int = 1000,
-  val roundExpiry: Duration = 1.seconds,
```
jframe, line 74: I think roundExpiry should be kept. Think is likely to be more constant than the block period, that is I don't think will change as the block period increases e.g. a block period of 2 seconds would also likely use a round expiry of 1 second. Unless a block period of 1 second is always used then it would be better to have this a seperate config parameter.

Filter94: > a block period of 2 seconds would also likely use a round expiry of 1 second

That would introduce an edge case bug, when the top transaction is taking more than 1 second to execute, and it also could be quite a realistic reason for a validator to skip its round. So I think the next block building attempt after the round expiry should also use the regular block time IMO.

Let me explain the reason I removed `roundExpiry`. If I were to roll it out on Linea, I'd set it to 2 seconds, because the block time right now is 2 seconds. If we change the block time to 1 second in the future, the old value of 2 seconds would introduce the unnecessary 1 second of delay until the resume of the block production in case there is a low load. So there is a room for us to forget setting this config correctly and if we just go with block time, we have a smaller room for error during deployment

jframe (approved):

PR #342, opened 2025-08-25
sets default syncing download config

fluentcrafter:


config/src/main/kotlin/maru/config/Config.kt
```diff
@@ -169,7 +169,7 @@ data class SyncingConfig(
   val syncTargetSelection: SyncTargetSelection,
   val elSyncStatusRefreshInterval: Duration,
   val desyncTolerance: ULong = 5UL,
-  val download: Download,
+  val download: Download = Download(),
```
Filter94, line 172: Why do you want to add it?

fluentcrafter: 1. convention over configuration
2. It was breaking the chaos deployments

Any reason not to add it?

Filter94: 1. Sorry, I don't catch what's the convention here

> Any reason not to add it?

Yes, the defaults were the reason why we missed the fact that the `DownloadBlocksStep`. Not these exact defaults, but rather the ones in `BeaconChainDownloadPipelineFactory`, so I got mad at defaults in general

Since these configs are deeply technical I think it's Ok to use defaults here, just wanted to clarify the reason

Filter94 (approved):

PR #345, opened 2025-08-26
On new fork, update forkId in Discovery Service

gauravahuja:


consensus/src/main/kotlin/maru/consensus/ProtocolStarter.kt
```diff
@@ -33,6 +35,7 @@ class ProtocolStarter(
       isDaemon,
     )
   },
+  private val forkTransitionSubscriptionManager: SubscriptionManager<ForkSpec>,
```
Filter94, line 38: We don't need to pass the manager here, because `ProtocolStarter` never adds new subscriptions there. It should just be a lambda with a callback

gauravahuja: ok, so I updated Subscription Manager that extends SubscriptionNotifier and Observable. Subscription Notifier is interface to only notify and not add new subscription.

app/src/main/kotlin/maru/app/MaruApp.kt
```diff
@@ -207,7 +210,8 @@ class MaruApp(
           finalizationStateProvider = finalizationProvider,
         )
       }
-
+    val forkTransitionSubscriptionManager = InOrderFanoutSubscriptionManager<ForkSpec>()
```
Filter94, line 213: I still don't understand why do we need `InOrderFanoutSubscriptionManager` instance here

consensus/src/test/kotlin/maru/consensus/ProtocolStarterTest.kt
```diff
@@ -72,9 +78,12 @@ class ProtocolStarterTest {
       createProtocolStarter(
         forksSchedule = forksSchedule,
         clockMilliseconds = 16000, // After fork transition at 15
+        forkTransitionNotifier = forkTransitionNotifier,
       )
     protocolStarter.start()
     assertActiveProtocol2(protocolStarter)
+    assertThat(forkTransitions.size).isEqualTo(1)
+    assertThat(forkTransitions.first()).isEqualTo(forkSpec2)
```
fluentcrafter, line 86: Simple assertions like this, imo, should be done as follow:
```suggestion
    assertThat(forkTransitions).isEqualTo(listOf(forkSpec2))
```

1. easier to read/understand
2. when it fails, having the list of items is easier to see what is missing or should not be in the list.

p2p/src/main/kotlin/maru/p2p/P2PNetworkImpl.kt
```diff
@@ -369,4 +373,17 @@ class P2PNetworkImpl(
...
     addStaticPeer(MultiaddrPeerAddress.fromAddress(address))
   }
+
+  override fun handleForkTransition(forkSpec: ForkSpec) {
+    val forkId =
+      ForkId(
+        chainId = chainId,
+        forkSpec = forkSpec,
+        genesisRootHash =
+          beaconChain.getBeaconState(0u)?.latestBeaconBlockHeader?.hash
```
fluentcrafter, line 383:
```
beaconChain.getBeaconState(0u)?.latestBeaconBlockHeader?
```
Probably unrelated to your PR, but this line is super controversial to me:
- 1. we get genesis block
- 2. from it, we get `latestBeaconBlockHeader`?

I think the problem steems from `BeaconState` data class, the `latestBeaconBlockHeader` should be renamed to `beaconBlockHeader`

@gauravahuja @Filter94 WDYT?

p2p/src/main/kotlin/maru/p2p/P2PNetworkImpl.kt
```diff
@@ -369,4 +373,17 @@ class P2PNetworkImpl(
...
+            ?: throw IllegalStateException("Genesis state not found"),
+      )
+    val newForkIdHash = forkIdHasher.hash(forkId)
+    discoveryService?.updateForkIdHash(Bytes.wrap(newForkIdHash))
+  }
```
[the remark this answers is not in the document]
fluentcrafter: This seems like a legit warning.

fluentcrafter (approved):

PR #349, opened 2025-08-27
chaos: improv wait target

fluentcrafter:


.github/workflows/main.yml
```diff
@@ -62,7 +62,7 @@ jobs:
 
   chaos-testing:
     needs: [ filter-commit-changes ]
-    if: ${{ always() && !cancelled() && needs.filter-commit-changes.outputs.has-changes-requiring-build == 'true' && github.ref == 'refs/heads/main' }}
```
Filter94, line 65: I wouldn't enable it on the feature branches just yet, given it's still flacky

chaos-testing/helm.mk
```diff
@@ -232,18 +232,59 @@ port-forward-restart-all-linea:
...
+			echo "Not all pods are running in default namespace. Waiting..."; \
+			sleep 2; \
+			continue; \
 		fi; \
-		running_pods=$$(kubectl get pods --namespace=default --field-selector=status.phase=Running --no-headers 2>/dev/null | wc -l | tr -d ' '); \
-		if [ "$$running_pods" -eq "$$total_pods" ]; then \
-			echo "All $$total_pods pods are running."; \
+		pods_ready=0; \
+		current_time=$$(date +%s); \
+		for line in $$pods_data; do \
```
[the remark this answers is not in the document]
Filter94: That's a good point. We emulate timeout logic with a `for` instead of `while true`

chaos-testing/helm.mk
```diff
@@ -232,18 +232,59 @@ port-forward-restart-all-linea:
 	$(MAKE) port-forward-linea
 
 wait-all-running:
```
Filter94, line 234: Can we move its contents to a shell script, broken down into functions? The way it is now, it's unreadable

.github/workflows/main.yml
fluentcrafter: The flakiness should be fixed now without restarting the bootnodes, which is good news :)

Also, I plan to refactor the CI as follows:
- main - run full chaos 
- PRs/branches - just deploy linea and check it spins up ok. Goal: validate that we refactor something we don't forget to update Chaos configs and find out in main
  - Optionally, allow to run full chaos by workflow dispatch

chaos-testing/helm.mk
fluentcrafter: Yes, I thought about that was well. Will do

chaos-testing/helm.mk
```diff
@@ -232,18 +232,59 @@ port-forward-restart-all-linea:
...
+	echo "Waiting for all pods in default namespace to be running for at least $$uptime_arg since last restart..."; \
+	uptime_seconds=0; \
+	case "$$uptime_arg" in \
+		*s) uptime_seconds=$$(echo "$$uptime_arg" | sed 's/s$$//'); ;; \
+		*m) uptime_seconds=$$(echo "$$uptime_arg" | sed 's/m$$//' | awk '{print $$1 * 60}'); ;; \
+		*h) uptime_seconds=$$(echo "$$uptime_arg" | sed 's/h$$//' | awk '{print $$1 * 3600}'); ;; \
+		*) echo "Invalid uptime format. Use format like: 30s, 2m, 1h"; exit 1; ;; \
+	esac; \
+	while true; do \
+	  total_pods=$$(kubectl get pods --namespace=default -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}' | wc -l | tr -d ' '); \
```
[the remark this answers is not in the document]
fluentcrafter: I had that, but it broke as i add/remove pods. This will an evolving thing.
So, it gets all pods and checks all are running for x seconds...

Issue #293, opened 2025-08-11
Fork change needs to change local node record


pinges:
When Maru is changing the Fork, the fork id hash on the local node record has to be updated.
Therefore, in MaruDiscoveryService, the method updateForkIdHash() needs to be called.

In addition the seq number of the local node record, which is created in createLocalNodeRecord(), needs to be:
- stored in the DB
- incremented when we are changing the local node record by calling updateForkIdHash()
- incremented when Maru starts

PR #353, opened 2025-08-27
Update discovery sequence number when enr changes

gauravahuja:
#293 

- Add interface for `RuntimeConfigs` in the Database package. `RuntimeConfigs` is used to get and update sequence number
- `KvDatabase` implements `BeaconChain` and `RuntimeConfigs`
- Inside discovery service Teku provides a callback option for localNodeRecord update. The localNodeRecord is updated whenever the fork id changes.
- In this callback I update the sequence number in the database.

### DIFF
--- a/core/src/main/kotlin/maru/database/RuntimeConfigs.kt
+++ b/core/src/main/kotlin/maru/database/RuntimeConfigs.kt
@@ -1,0 +1,23 @@
+/*
+ * Copyright Consensys Software Inc.
+ *
+ * This file is dual-licensed under either the MIT license or Apache License 2.0.
+ * See the LICENSE-MIT and LICENSE-APACHE files in the repository root for details.
+ *
+ * SPDX-License-Identifier: MIT OR Apache-2.0
+ */
+package maru.database
+
+interface RuntimeConfigs : AutoCloseable {
+  fun getDiscoverySequenceNumber(): ULong
+
+  fun newRuntimeConfigsUpdater(): Updater
+
+  interface Updater : AutoCloseable {
+    fun putDiscoverySequenceNumber(newSequenceNumber: ULong): Updater
+
+    fun commit(): Unit
+
+    fun rollback(): Unit
+  }
+}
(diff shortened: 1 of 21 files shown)


fluentcrafter (commented):
LGTM overall, just a remark on the interface name as I think it may be a little misleading

p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
```diff
@@ -202,10 +207,11 @@ class MaruDiscoveryService(
   }
 
   private fun createLocalNodeRecord(): NodeRecord {
+    val sequenceNumber = runtimeConfigs.getDiscoverySequenceNumber() + 1uL
```
pinges, line 210: The localNodeRecordUpdated is only called when the record is changed, not when a new LocalNodeRecordStore is created. That means we have to store that new sequence number here as well.

pinges (commented):
Another thing that is missing is that when we call the updateForkIdHash in MaruDiscoveryService we need to update the sequence number as well, as that is not done automatically.

core/src/main/kotlin/maru/database/RuntimeConfigs.kt
```diff
@@ -0,0 +1,23 @@
...
+ * Copyright Consensys Software Inc.
+ *
+ * This file is dual-licensed under either the MIT license or Apache License 2.0.
+ * See the LICENSE-MIT and LICENSE-APACHE files in the repository root for details.
+ *
+ * SPDX-License-Identifier: MIT OR Apache-2.0
+ */
+package maru.database
+
+interface RuntimeConfigs : AutoCloseable {
```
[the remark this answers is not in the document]
Filter94: `ServiceState` as an option

p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
```diff
@@ -121,6 +125,7 @@ class MaruDiscoveryService(
       .listen(p2pConfig.ipAddress, p2pConfig.discovery!!.port.toInt())
       .secretKey(privateKey)
       .localNodeRecord(createLocalNodeRecord())
+      .localNodeRecordListener(this::localNodeRecordUpdated)
```
Filter94, line 128: Nit: can we omit `this` here?

fluentcrafter: tbh, i kind of like the this:: because it makes it clean if it is an instance method vs a static reference from somewhere else.

fluentcrafter (approved):

PR #357, opened 2025-08-29
Switch to paris

Filter94:
* Implemented an ability to switch from Clique to Paris based on TTD
* Changed Long to ULong in `ForkSpec` and `ForksSchedule` and as a consequence, almost everywhere on the repo
* Modified `CliqueToPosTest` and `MaruConsensusSwitchTest` to verify that some Paris blocks are produced successfully
* Reimplemented `ElDelegatedConsensus` which now tracks blocks difficulty and if TTD is reached, it runs the QBFT protocol with Prague instead
* Added `ParisWeb3JJsonRpcExecutionLayerEngineApiClient`
* Reimplemented `FollowerBeaconBlockImporterTest`. The reason is that mocks that were used are incompatible with Kotlin's unsigned type. I replaced the mocks with test doubles
* Removed an exception throw from `ChainDataProvider`, because `getBeaconStateByStateRoot` returns non-nullable type


app/src/main/kotlin/maru/app/MaruApp.kt
```diff
@@ -215,12 +216,18 @@ class MaruApp(
       }
     val forkTransitionSubscriptionManager = InOrderFanoutSubscriptionManager<ForkSpec>()
     forkTransitionSubscriptionManager.addSyncSubscriber(p2pNetwork::handleForkTransition)
+    val elDelegatedConsensusFactory =
+      ElDelegatedConsensusFactory(
+        ethereumJsonRpcClient = validatorELNodeEthJsonRpcClient.eth1Web3j,
```
jonesho, line 221: not related to logic: but why here we called it `validatorELNodeEthJsonRpcClient.eth1Web3j` where the El node could be from follower

app/src/main/kotlin/maru/app/MaruAppFactory.kt
```diff
@@ -121,14 +121,7 @@ class MaruAppFactory {
         }
 
     val qbftFork = beaconGenesisConfig.getForkByConfigType(QbftConsensusConfig::class)
-    val qbftForkTimestamp = qbftFork.timestampSeconds.toULong()
     val qbftConfig = qbftFork.configuration as QbftConsensusConfig
-    BeaconChainInitialization(
```
jonesho, line 126: This constructor was mis-placed before given that it would be called before at line 120?

serialization/src/main/kotlin/maru/serialization/ForkIdSerializers.kt
```diff
@@ -40,17 +40,17 @@ object ForkIdSerializers {
...
             .array()
         }
 
         is ElDelegatedConfig -> {
           ByteBuffer
             .allocate(4 + 8 + 1)
-            .putInt(value.blockTimeSeconds)
-            .putLong(value.timestampSeconds)
+            .putInt(value.blockTimeSeconds.toInt())
+            .putLong(value.timestampSeconds.toLong())
```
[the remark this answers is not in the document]
Filter94: Good point. I missed that

jvm-libs/mappers/src/main/kotlin/maru/mappers/Mappers.kt
```diff
@@ -153,7 +154,7 @@ object Mappers {
       transactions = this.transactions.map { it.toArray() },
     )
 
-  fun ExecutionPayloadV2.toDomainExecutionPayload() =
+  fun ExecutionPayloadV1.toDomainExecutionPayload() =
```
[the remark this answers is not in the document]
Filter94: Since Shanghai is already implemented, I don't want to remove it just yet

app/src/main/kotlin/maru/app/MaruAppFactory.kt
```diff
@@ -381,14 +374,9 @@ class MaruAppFactory {
...
   ) {
     val qbftForkConfig = beaconGenesisConfig.getForkByConfigType(QbftConsensusConfig::class)
-    val qbftForkTimestamp =
-      qbftForkConfig
-        .timestampSeconds
-        .toULong()
     val beaconChainInitialization =
       BeaconChainInitialization(
         beaconChain = beaconChain,
-        genesisTimestamp = qbftForkTimestamp,
```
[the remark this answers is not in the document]
Filter94: I forgor, but there was some issue related to that

docker/initialization/genesis-maru.json.template
```diff
@@ -3,7 +3,12 @@
   "config": {
     "0": {
       "type": "delegated",
-      "blockTimeSeconds": 1
+      "blockTimeSeconds": 4,
```
jonesho, line 6: Why `blockTimeSeconds` = 4 instead of 1 here?

consensus/src/main/kotlin/maru/consensus/delegated/ElDelegatedConsensus.kt
```diff
@@ -8,113 +8,153 @@
...
+  @Synchronized
+  private fun transitionToPostTtdProtocol(postTtdForkSpec: ForkSpec) {
+    if (postTtdProtocol != null) {
+      throw IllegalStateException("This protocol is supposed to be stopped after reaching TTD")
+    }
+
+    try {
+      log.info("Creating post-TTD protocol forkSpec={}", postTtdForkSpec)
+      postTtdProtocol = postTtdProtocolFactory.create(postTtdForkSpec)
+      postTtdProtocol?.start()
```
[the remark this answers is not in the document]
Filter94: The main motivation for that is to make it fit into the existing architecture and to avoid the redesign. A good way to redesign it would be, as you suggested, is to make the `ForksScheduleAdapter` work properly based on timestamps, rather than block numbers as it is now. But that requires changes in QBFT core and we don't have time to change it right now

app/src/main/kotlin/maru/app/MaruApp.kt
Filter94: We need a node on which we have the most control and which provides the strongest consistency with Maru's beacon chain. That's the payload validator el node. Also, that's the only mandatory node. Followers list may be empty

app/src/main/kotlin/maru/app/MaruAppFactory.kt
Filter94: That's my understanding, yes. I think we were trying to initialized the Beacon chain 2 times, but the second time it was just making sure that the state is initialized

docker/initialization/genesis-maru.json.template
Filter94: There is no reason actually. And even for Maru, I think we've fixed an issue because of which we've set 7 seconds of block time. So I'll try to set it to 1 second and will see how it works on CI

app/src/main/kotlin/maru/app/MaruAppFactory.kt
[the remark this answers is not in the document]
Filter94: Now I recall and it's causing an issue in Devnet. This thing was set to the first QBFT timestamp, which isn't the first anymore. The first timestamp should be either Paris switch, which we don't know in advance, or something else before that. I chose 0 as the safest option

PR #361, opened 2025-09-02
feat: update erigon version to v3.0.2

jonesho:


docker/erigon/config.toml
```diff
@@ -0,0 +1,16 @@
+"datadir" = "/home/erigon"
+"networkid" = "1337"
+"miner.gaslimit" = "0x1C9C380"
+"http" = true
+"http.addr" = "0.0.0.0"
```
Filter94, line 5: Can we get rid of the quotes and use sections instead?

jonesho: I tried to use sections like what we did in Geth config.toml but it didn't work, seems like erigon requires this
```
"miner.gaslimit"  = "0x1C9C380"
```
and rejects
```
[miner]
gaslimit = "0x1C9C380"
```

Filter94 (approved):

keithchew:
Hi @jonesho 

I am trying to get Erigon v3.3.0 working with Linea, as that contains the Fusaka fork updates. Do you know what are the specific Prague errors you were getting when you had to pin it to v3.0.2? Looks like I might be hitting some errors which I am happy to help investigate and fix.

PR #362, opened 2025-09-03
Increment sequence number in db on initializaiton of discovery service

gauravahuja:


p2p/src/main/kotlin/maru/p2p/discovery/MaruDiscoveryService.kt
```diff
@@ -208,6 +208,10 @@ class MaruDiscoveryService(
 
   private fun createLocalNodeRecord(): NodeRecord {
     val sequenceNumber = p2PState.getLocalNodeRecordSequenceNumber() + 1uL
+    p2PState
+      .newP2PStateUpdater()
+      .putDiscoverySequenceNumber(sequenceNumber)
+      .commit()
```
Filter94, line 214: Can we please remove the TODO below then?

Filter94 (approved):

Issue #325, opened 2025-08-21
Additional App tests


Filter94:
## Description
There are a few scenarios that we want to cover as App tests

## Motivation
These are realistic scenarios that can happen in Production, so we better cover them as tests rather than catch bugs in Production 

## Tasks
- [ ] T1 - Implement a test where Sequencer Maru is restarted from scratch and it's able to sync its state from peers and continue to produce blocks
- [x] T2 - Implement a test where Sequencer Besu is restarted from scratch and it's able to sync its state from peers and continue to produce blocks
- [x] T3 - Implement a test where there is 1 Sequencer stack and 1 follower stack. Transactions are sent 1 by one through the follower stack. When Validator Stack is interrupted (Maru or Besu) block production should resume eventually
- [x] T4 - Implement a test where there is 1 Sequencer stack and 1 follower stack with 1 Besu Payload Validator and 1 Besu Follower node. This stack is started with a delay to imitate a sync from scratch. Transactions are sent 1 by 1 directly to Follower. Maru follower is able to sync without getting out of sync, Besu Follower receives updates and is able to follow the chain

## Acceptance criteria
- [ ] The mentioned scenarios are covered

## Risks
- [ ] If these aren't covered, we may hit these issues in Production

## Remember to
 - [ ] Add the `documentation` label in case there is an impact on the documentation
 - [ ] Add `priority` and `team` labels
 - [ ] Add Task for updating the Runbook or adding/updating existing metrics and alerts.

fluentcrafter:
@Filter94 T1 and T5 sound the same to me in different words. Can you please elaborate a little more?

Filter94:
@fluentcrafter I think it's just a duplicate. I'll remove it

PR #376, opened 2025-09-16
[325] Implement test for delayed start follower (t4)

gauravahuja:
#325

### DIFF
--- a/app/src/integrationTest/kotlin/maru/app/MaruFollowerDelayedStartTest.kt
+++ b/app/src/integrationTest/kotlin/maru/app/MaruFollowerDelayedStartTest.kt
@@ -1,0 +1,112 @@
+/*
+ * Copyright Consensys Software Inc.
+ *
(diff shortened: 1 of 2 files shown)


app/src/integrationTest/kotlin/maru/app/MaruFollowerDelayedStartTest.kt
```diff
@@ -0,0 +1,112 @@
...
+          recipient = createAccount("another account"),
+          amount = Amount.ether(100),
+        )
+      }
+    }
+
+    validatorStack.besuNode.assertMinedBlocks(2 * blocksToProduce)
+    followerStack.besuNode.assertMinedBlocks(2 * blocksToProduce)
+
+    followerStack.maruApp.stop()
```
Filter94, line 107: It's better to move them into a `@AfterEach`, because in case this test fails, the nodes may be left hanging

PR #382, opened 2025-09-22
Added filter to return only the connected peers

Filter94:


p2p/src/main/kotlin/maru/p2p/MaruReputationManager.kt
```diff
@@ -142,7 +142,7 @@ class MaruReputationManager(
       reason: Optional<DisconnectReason>,
       locallyInitiated: Boolean,
     ) {
-      if (suitableAfter.isEmpty) {
+      if (isSuitableAt(disconnectTime)) {
```
Filter94, line 145: This fixes the duplicate ban bug

p2p/src/main/kotlin/maru/p2p/MaruPeerManager.kt
```diff
@@ -159,9 +149,14 @@ class MaruPeerManager(
     log.trace("Peer={} disconnected", peer.id)
   }
 
-  override fun getPeer(nodeId: NodeId): MaruPeer? = connectedPeers[nodeId]
+  private fun actuallyConnectedPeers(): Map<NodeId, MaruPeer> = connectedPeers.filter { it.value.isConnected }
```
Filter94, line 152: This fixes a bug when the sync pipeline tries to sync from a disconnected peer

PR #395, opened 2025-09-30
Optional payload validation

Filter94:
> [!NOTE]
> Adds a config flag to enable/disable execution payload validation and wires it through factories, validators, and tests.
> 
> - **Consensus/Validation**:
>   - `BeaconBlockValidatorFactoryImpl`: accept nullable `executionLayerManager`; only include `ExecutionPayloadValidator` when provided; minor refactor for empty-block validator.
>   - `QbftValidatorFactory` and `QbftFollowerFactory`: accept `payloadValidationEnabled`; pass `null` `executionLayerManager` to validator factory when disabled.
> - **App Wiring**:
>   - `MaruApp`: propagate `config.validatorElNode.payloadValidationEnabled` into protocol factories.
> - **Config**:
>   - `ValidatorElNode`: add `payloadValidationEnabled`.
>   - `PayloadValidatorDto`: add `payloadValidationEnabled` (default `true`) and map to domain.
> - **Tests**:
>   - Add integration test case: follower imports blocks with payload validation disabled.
>   - Update Hoplite parsing tests; add cases for explicit `true` and default enabled; expectations updated to include the flag.
> - **Test Utilities**:
>   - `MaruFactory`: plumb `enablePayloadValidation` through config builders and follower creation helpers.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 821c15a1a127a28dbf9e81b81320d09f8792d7dc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/jvm-libs/test-utils/src/main/kotlin/testutils/maru/MaruFactory.kt
+++ b/jvm-libs/test-utils/src/main/kotlin/testutils/maru/MaruFactory.kt
@@ -602,6 +602,7 @@ class MaruFactory(
         p2pConfig = p2pConfig,
         overridingLineaContractClient = overridingLineaContractClient,
         syncingConfig = syncingConfig,
+        enablePayloadValidation = enablePayloadValidation,
       )
     return buildApp(
       config,


config/src/test/kotlin/maru/config/HopliteFriendlinessTest.kt
```diff
@@ -64,6 +64,7 @@ class HopliteFriendlinessTest {
     [payload-validator]
     engine-api-endpoint = { endpoint = "http://localhost:8555", jwt-secret-path = "/secret/path" }
     eth-api-endpoint = { endpoint = "http://localhost:8545" }
+    payload-validation-enabled = false
```
fluentcrafter, line 67: IIUC, this only takes effect when we get block from P2P network, we skip it's validation.

Or, if we don't do payload validation we don't need `[payload-validator]` section at all?

Reason asking is that it would be good to find a way of running maru without `[payload-validator]` this would allow to easily:
- Run a Maru Bootnode
- Run a Maru for Snapshot taking 
- Sync Maru locally with a prod network that needs to have full EL state. 

I know it's not your intent of this PR. But in case it's very close, maybe worth the extra stretch ?

fluentcrafter (approved):

PR #413, opened 2025-10-08
refactor tests to avoid repeated code

fluentcrafter:


core/src/testFixtures/kotlin/maru/consensus/ForkIdManagerFactory.kt
```diff
@@ -0,0 +1,69 @@
...
+    consensusConfig: ConsensusConfig =
+      QbftConsensusConfig(
+        validatorSet =
+          setOf(
+            DataGenerators.randomValidator(),
+            DataGenerators.randomValidator(),
+          ),
+        elFork = elFork,
+      ),
+    forks: List<ForkSpec> = listOf(ForkSpec(0UL, 1u, consensusConfig)),
```
Filter94, line 30: Why don't we use the `ForksSchedule` here directly?

PR #428, opened 2025-10-14
Add config option for dns caching, default: no caching

gauravahuja:
From https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/net/doc-files/net-properties.html

> Since these 3 properties are part of the security policy, they are not set by either the -D option or the System.setProperty() API, instead they are set as security properties.
> 
> The three properties that control dns cache are:
> - networkaddress.cache.ttl
> - networkaddress.cache.stale.ttl
> - networkaddress.cache.negative.ttl
> 

Default values for networkaddress.cache.ttl and networkaddress.cache.negative.ttl is set to zero ie no caching.

---

> [!NOTE]
> Adds `networkAddressCacheTTLSeconds` and `networkAddressCacheNegativeTTLSeconds` config options (default 0) and sets corresponding Java DNS cache security properties at startup; updates TOML DTO/mapping and tests.
> 
> - **Config**:
>   - Add `MaruConfig` fields: `networkAddressCacheTTLSeconds`, `networkAddressCacheNegativeTTLSeconds` (defaults 0).
>   - Extend `MaruConfigDtoToml` with matching fields and map them in `domainFriendly()`.
> - **App Startup**:
>   - Set Java security properties `networkaddress.cache.ttl` and `networkaddress.cache.negative.ttl` from config in `MaruAppFactory`.
> - **Tests**:
>   - Update Hoplite tests to include and assert the new fields in TOML parsing and domain conversion.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f7c068b8939855c0833b445d968e366fe3b0defa. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/config/src/main/kotlin/maru/config/Config.kt
+++ b/config/src/main/kotlin/maru/config/Config.kt
@@ -237,6 +237,8 @@ data class SyncingConfig(
 }
 
 data class MaruConfig(
+  val networkAddressCacheTTLSeconds: Int = 0,
+  val networkAddressCacheNegativeTTLSeconds: Int = 0,
   val protocolTransitionPollingInterval: Duration = 1.seconds,
   val allowEmptyBlocks: Boolean = false,
   val persistence: Persistence,
--- a/app/src/main/kotlin/maru/app/MaruAppFactory.kt
+++ b/app/src/main/kotlin/maru/app/MaruAppFactory.kt
@@ -101,6 +101,16 @@ class MaruAppFactory {
       () -> SyncStatusProvider,
     ) -> P2PNetworkImpl = ::P2PNetworkImpl,
   ): MaruApp {
+    java.security.Security.setProperty(
+      "networkaddress.cache.ttl",
+      config.networkAddressCacheTTLSeconds.toString(),
+    )
+
+    java.security.Security.setProperty(
+      "networkaddress.cache.negative.ttl",
+      config.networkAddressCacheNegativeTTLSeconds.toString(),
+    )
+
     log.info("configs={}", config)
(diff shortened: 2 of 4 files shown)


config/src/main/kotlin/maru/config/Config.kt
```diff
@@ -237,6 +237,8 @@ data class SyncingConfig(
 }
 
 data class MaruConfig(
+  val networkAddressCacheTTLSeconds: Int = 0,
+  val networkAddressCacheNegativeTTLSeconds: Int = 0,
```
fluentcrafter: I don't think this "one size fits" all is a good approach. 
The current implementation is **always** overriding the default DNS config. 

However, this can be problematic if Maru runs on a different JVM without access to the security (common on enterprizes with high security constraints), where the MaruAppFactory would blow when calling setProperty.

From [Java doc](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/security/Security.html#setProperty(java.lang.String,java.lang.String)) we can see

> First, if there is a security manager, its checkPermission method is called with a java.security.SecurityPermission("setProperty."+key) permission to see if it's ok to set the specified security property value.

My proposal is to:
- make these configs optional (`Int? = null`) with default to null
- only call setProperty when it's defined..

gauravahuja: > always overriding the default DNS config

This is only for the application. Not for any other process that runs on jvm on the node.

Issue #407, opened 2025-10-07
Add builtin network configuration


jframe:
## Description
Add builtin network configurations for Linea mainnet and Sepolia testnet to Maru. This would allow users to easily connect to these networks without manually configuring the genesis file. Users would be able to start Maru with a simple `--network=linea-mainnet` or `--network=linea-sepolia` flag instead of providing custom configuration files.

## Motivation
- **Improved user experience**: Reduces friction for users wanting to test Maru on public networks
- **Reduced configuration errors**: Pre-validated configs eliminate common setup mistakes
- **Faster onboarding**: New users can quickly spin up a node without hunting for correct network parameters

## Tasks
- [ ] Create CLI flag/arguments (e.g., `--network=linea-mainnet|linea-sepolia`)
- [ ] Add network configuration for Linea Sepolia and Linea Mainnet networks
- [ ] Add network configuration loading logic

## Acceptance criteria
- [ ] User can start Maru with `--network=linea-mainnet` and connect to Linea mainnet
- [ ] User can start Maru with `--network=linea-sepolia` and connect to Linea Sepolia testnet

## Remember to
- [ ] Add the `documentation` label in case there is an impact on the documentation
- [ ] Add `priority` and `team` labels
- [ ] Add Task for updating the Runbook or adding/updating existing metrics and alerts

PR #451, opened 2025-10-27
Making Maru work without an L2 EL node

Filter94:
Fixes #442 

---

> [!NOTE]
> Allows Maru to operate without a local EL by making EL config optional, introducing `fork-transition.l2-eth-api-endpoint`, and updating factories/sync to work when EL is absent, with new tests and Helm/docker configs.
> 
> - **Config**:
>   - Make `validatorElNode` optional; add `fork-transition` with `l2-eth-api-endpoint` and `protocol-transition-polling-interval`.
>   - Rename Linea fields to `l1EthApiEndpoint` and add required `l2EthApiEndpoint`; DTO/back-compat and validations updated.
>   - Update TOML loaders, defaults, and tests accordingly.
> - **App/Factory**:
>   - Accept nullable EL clients; expose `apiPort()`; close clients conditionally.
>   - Create EL manager map only when validator EL is configured; EL sync becomes optional via `NoOpLongRunningService`.
>   - Add `checkL2EthApiEndpointAndForks` and wire `forkTransition` interval.
> - **Consensus/Sync**:
>   - `DifficultyAwareQbftFactory` takes optional Web3j; require presence on create; `ProtocolStarter` uses `onBeaconSyncComplete`.
>   - `SyncController.create` refactored to accept `elSyncServiceFactory` instead of concrete handlers.
> - **Follower without EL**:
>   - `QbftFollowerFactory` imports payload to EL only if client exists and validates conditionally.
>   - New integration test `MaruFollowerNoElTest` verifies block import without EL.
> - **Helm/Docker**:
>   - Move EL JSON-RPC URLs to `[fork-transition]`; remove `eth-api-endpoint` from `[payload-validator]`.
>   - Update chaos-testing values and `docker/maru/config.dev.toml` accordingly.
> - **Build/CI**:
>   - Add test dependency in `app/build.gradle`; fix `p2p` testFixtures; bump `azure/setup-helm` to `v4.3.1`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit edae720db0424e241058572fe14d92d8b96cf6f6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #453, opened 2025-10-30
feat: added builtin network configuration for Linea sepolia and Linea…

jonesho:
This implements https://github.com/Consensys/maru/issues/407

To ensure backward-compatibility, we require either one of `--maru-genesis-file` and `--network` needs to be specified.

---

> [!NOTE]
> Adds --network support with built-in Linea mainnet/sepolia genesis files, defaulting to mainnet if unspecified, plus CLI parsing tweaks and tests.
> 
> - **CLI**:
>   - Add `--network` (values: `linea-mainnet`, `linea-sepolia`) via `Network` enum; default to `linea-mainnet` when neither `--genesis-file` nor `--network` is provided.
>   - Introduce `GenesisOptions` arg group (mutually exclusive `--genesis-file`/`--maru-genesis-file` vs `--network`).
>   - Add `KebabToEnumConverter` and register in `CliEntrypoint`; supports case-insensitive kebab-case.
>   - Switch file validation to `validateFileCanRead`; add `dryRun` to skip app startup and relax file checks.
>   - Replace stderr with structured logging for errors.
> - **Config Loader**:
>   - Change `loadGenesisConfig` to accept a `String` path instead of `File`.
> - **Resources**:
>   - Add built-in genesis files: `app/src/main/resources/beacon-genesis-files/linea-mainnet-genesis.json` and `.../linea-sepolia-genesis.json`.
> - **Tests**:
>   - Add `MaruAppCliTest` covering parsing, defaults, exclusivity rules, invalid inputs, and case-insensitivity.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e33c183b1cedf9dab3a2fd8e93f229a579319697. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/src/main/kotlin/maru/app/MaruAppCli.kt
+++ b/app/src/main/kotlin/maru/app/MaruAppCli.kt
@@ -60,6 +60,8 @@ enum class Network(
 class MaruAppCli(
   private val dryRun: Boolean = false,
 ) : Callable<Int> {
+  private val log = LogManager.getLogger(this.javaClass)
+
   @Option(
     names = ["--config"],
     paramLabel = "CONFIG.toml,CONFIG.overrides.toml",
@@ -94,7 +96,7 @@ class MaruAppCli(
   override fun call(): Int {
     for (configFile in configFiles!!) {
       if (!validateFileCanRead(configFile)) {
-        System.err.println("Failed to read config file: \"${configFile.path}\"")
+        log.error("Failed to read config file: \"${configFile.path}\"")
         return 1
       }
     }
@@ -105,13 +107,13 @@ class MaruAppCli(
     }
     if (genesisOptions!!.genesisFile != null) {
       if (!validateFileCanRead(File(genesisOptions!!.genesisFile!!))) {
-        System.err.println("Failed to read genesis file: \"${genesisOptions!!.genesisFile}\"")
+        log.error("Failed to read genesis file: \"${genesisOptions!!.genesisFile}\"")
         return 1
       }
-      println("Using the given genesis file from \"${genesisOptions!!.genesisFile}\"")
+      log.info("Using the given genesis file from \"${genesisOptions!!.genesisFile}\"")
     } else {
       genesisOptions!!.genesisFile = buildInGenesisFileResourcePath(genesisOptions!!.network!!.networkNameInKebab)
-      println("Using the genesis file of the named network \"${genesisOptions!!.network!!.networkNameInKebab}\"")
+      log.info("Using the genesis file of the named network \"${genesisOptions!!.network!!.networkNameInKebab}\"")
     }
 
     if (!dryRun) {


app/src/main/kotlin/maru/app/MaruAppCli.kt
```diff
@@ -53,12 +70,23 @@ class MaruAppCli : Callable<Int> {
...
-    val parsedBeaconGenesisConfig = MaruConfigLoader.loadGenesisConfig(genesisFile)
+
+    if (genesisOptions.genesisFile != null) {
+      if (!validateConfigFile(genesisOptions.genesisFile!!)) {
+        System.err.println("Failed to read genesis file file: \"${genesisOptions.genesisFile!!.path}\"")
+        return 1
+      }
+      println("Using the given genesis file from \"${genesisOptions.genesisFile!!.path}\"")
+    } else {
+      println("Using the default genesis file from Linea ${genesisOptions.network}")
```
Filter94:
```suggestion
      println("Using the genesis file of the named network ${genesisOptions.network}")
```

app/src/main/kotlin/maru/app/MaruAppCli.kt
```diff
@@ -36,29 +53,57 @@ class MaruAppCli : Callable<Int> {
...
-    names = ["--maru-genesis-file"],
-    paramLabel = "BEACON_GENESIS.json",
-    description = ["Beacon chain genesis file"],
-    required = true,
+  @CommandLine.ArgGroup(multiplicity = "1", exclusive = true, validate = true)
+  lateinit var genesisOptions: GenesisOptions
+
+  class GenesisOptions(
+    @CommandLine.Option(
+      names = ["--maru-genesis-file"],
```
jframe: I think this can be just `--genesis-file` no need for the maru prefix

jonesho: I'd not recommend this for now, as this would break backward-compatibilty for our external partners, unless we really want to do it and have a plan to communicate clearly to current external partners

jframe: Fair enough, better to leave as is then

Filter94: I think we can take take the same strategy as with linea.l1EthRpc [here](https://github.com/Consensys/maru/pull/451#discussion_r2466067069)

Meaning that we'd accept both `maru-genesis-file` and `genesis-file` and maybe even print a warning that the former is deprecated

app/src/main/kotlin/maru/app/MaruAppCli.kt
```diff
@@ -17,6 +17,23 @@ import org.apache.logging.log4j.core.config.Configurator
...
 import picocli.CommandLine.Command
 
+internal class KebabToEnumConverter<T : Enum<T>>(
+  private val enumClass: Class<T>,
+) : CommandLine.ITypeConverter<T> {
+  override fun convert(value: String): T {
+    // Convert kebab-case to upper snake-case
+    val enumName = value.replace('-', '_').uppercase()
+    return try {
+      java.lang.Enum.valueOf(enumClass, enumName)
```
Filter94: Minor: Can you please add an import instead?

app/src/main/kotlin/maru/app/MaruAppCli.kt
Filter94: How can we test this change? I think now we're in position to invest some time to make this class testable without instantiating the whole application

app/src/main/kotlin/maru/app/MaruAppCli.kt
jonesho: I will leave it as is, cause import `java.lang.Enum` would conflict with the Kotlin `Enum` used in line 20

app/src/main/kotlin/maru/app/MaruAppCli.kt
```diff
@@ -53,12 +91,24 @@ class MaruAppCli : Callable<Int> {
...
+    }
+    if (genesisOptions!!.genesisFile != null) {
+      if (!validateConfigFile(genesisOptions!!.genesisFile!!)) {
+        System.err.println("Failed to read genesis file file: \"${genesisOptions!!.genesisFile!!.path}\"")
+        return 1
+      }
+      println("Using the given genesis file from \"${genesisOptions!!.genesisFile!!.path}\"")
+    } else {
+      genesisOptions!!.genesisFile =
+        File("/beacon-genesis-files/${genesisOptions!!.network!!.networkNameInKebab}-genesis.json")
```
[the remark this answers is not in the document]
Filter94: It's supposed to load them from jar's resources, right? I don't see how `File("/beacon-genesis-files...")` can work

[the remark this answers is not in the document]
Filter94: Looking at the [source code](https://github.com/sksamuel/hoplite/blob/master/hoplite-core/src/main/kotlin/com/sksamuel/hoplite/ConfigSource.kt#L69) itself, they're using Path to check if it exists on the FS and then they try to load the config from the resources.

Since there is no class in Java generalizing resources and files, I suggest to go with a plain String, that doesn't imply anything on where the genesis file is located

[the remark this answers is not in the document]
Filter94: One more thing I realized. It still makes sense to do the old file validation `validateConfigFile` in case the genesis file is specified in `genesisFile`, so please keep that functionality

app/src/main/kotlin/maru/app/MaruAppCli.kt
```diff
@@ -28,61 +57,90 @@ import picocli.CommandLine.Command
   optionListHeading = "%nOptions:%n",
   footerHeading = "%n",
 )
-class MaruAppCli : Callable<Int> {
-  @CommandLine.Option(
+class MaruAppCli(
+  private val dryRun: Boolean = false,
```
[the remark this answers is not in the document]
Filter94: I can agree that `MaruConfigLoader` can have its own tests, but it doesn't cancel the fact that `MaruAppCli` still has some logic to bind stuff together, which at the moment is not testable

I added this ticket as a follow-up

app/src/main/kotlin/maru/app/MaruAppCli.kt
```diff
@@ -28,61 +57,92 @@ import picocli.CommandLine.Command
...
+            }
+          },
+        )
+    }
 
     return 0
   }
 
-  private fun validateConfigFile(file: File): Boolean = file.canRead()
+  private fun validateFileCanRead(file: File): Boolean = dryRun || file.canRead()
```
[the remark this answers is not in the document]
jonesho: Lets do it on this follow-up ticket: https://github.com/Consensys/maru/issues/456

PR #455, opened 2025-10-31
Maru testing cluster

fluentcrafter:
> [!NOTE]
> Introduce reusable Maru/Besu test cluster libs and genesis builders, add comprehensive tests, and fix P2P discovery port and fork selection with small API/cleanup tweaks.
> 
> - **Test Utilities (new module `jvm-libs/test-utils`)**:
>   - **Clusters**: Add `BesuCluster` and `MaruCluster` for programmatic node orchestration (start/stop, bootnodes, dynamic add/remove, discovery/static peering).
>   - **Config/Genesis**: Add helpers (`MaruConfigHelper`, `GenesisFactory`, `MaruGenesisFactory`, `BesuGenesisFactory`) and bundled `besu-genesis-template.json`.
>   - **Factories/Builders**: `NodeBuilder`, `TestMaruAppFactory.createMaru`, `BesuFactory` enhancements (naming, discovery, sync config).
>   - **Extensions**: Helpers for `BesuNode`, `MaruApp`, `MaruCluster`; `NetworkUtil` for free ports.
>   - **Tests**: Add `BesuClusterTest`, `MaruClusterTest`, and genesis factory tests.
> - **P2P**:
>   - Use actual listening port for discovery when configured as `0`; expose `nodeAddresses` via getter; improve fork selection in `ForkPeeringManager` when first fork is in the future.
>   - Test updates: replace `TestUtils.findFreePort` with new `NetworkUtil`; parameterize ports.
> - **Core/API**:
>   - `BeaconChain`: add `getLatestBeaconBlock()` convenience.
> - **App**:
>   - `MaruApp.close()`: close `protocolStarter` before DB; minor directory/private key handling refactor in `MaruAppFactory`.
> - **Build/Config**:
>   - Opt-in to `kotlin.time.ExperimentalTime`; update test-utils dependencies; enable parallel test forks.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 47ffb935545806d561adac22a6f2f42ed08067bd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


jvm-libs/test-utils/src/main/kotlin/maru/test/cluster/MaruCluster.kt
```diff
@@ -0,0 +1,313 @@
...
+    }
+    besuCluster.start(awaitPeerDiscovery = besuClusterWaitPeerDiscovery)
+    println("\n\n=== Besu cluster started ===\n\n")
+  }
+
+  private fun getBootnodesEnrs(): List<String> =
+    nodes
+      .filter {
+        it.nodeRole.isBootnode()
+      }.map { it.maru.p2pNetwork.enr!! }
```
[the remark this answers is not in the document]
fluentcrafter: Not an issues as bootnodes need to have enr defined.

app/src/main/kotlin/maru/app/MaruAppFactory.kt
```diff
@@ -405,12 +405,14 @@ class MaruAppFactory {
         log.info("Maru is using private key defined in file={}", privateKeyPath.toString())
       }
 
+      ensureDirectoryExists(privateKeyPath.parent)
```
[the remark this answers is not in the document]
fluentcrafter: I am not getting the point. Do you want me to remove it, if so why exactly?

jvm-libs/test-utils/src/main/kotlin/maru/test/cluster/MaruCluster.kt
```diff
@@ -0,0 +1,313 @@
...
+
+  fun start(): MaruCluster {
+    if (runningState != RunningState.STOPPED) {
+      throw IllegalStateException("Cannot START MaruCluster that is $runningState")
+    }
+    this.runningState = RunningState.STARTING
+    val nodesStartConfigs = nodesBuilders.map { it.build(genesisFactory::besuGenesis, besuCluster) }
+
+    val sequencersAddress =
```
Filter94: Addresses

jvm-libs/test-utils/src/main/kotlin/maru/test/cluster/MaruCluster.kt
```diff
@@ -0,0 +1,313 @@
...
+      .forEach { nodeStartingConfig ->
+        nodes += buildClusterNode(nodeStartingConfig, bootNodesEnrs)
+      }
+
+    this.runningState = RunningState.RUNNING
+    nodesBuilders.clear()
+    return this
+  }
+
+  fun buildClusterNode(
```
Filter94: Is this exposed for the negative tests? I believe normally, `addNewExtraNodeAndStart` should be enough for users

jvm-libs/test-utils/src/main/kotlin/maru/test/cluster/NodeBuilder.kt
```diff
@@ -0,0 +1,137 @@
...
+
+  fun maruConfig(function: (MaruConfig) -> MaruConfig) {
+    this.maruConfig = function(maruConfig)
+  }
+
+  private fun initBesuBuilderFunctionIfNecessary(
+    besuGenesisProvider: () -> String,
+    besuCluster: BesuCluster,
+  ) {
+    if (elNodeBuilder == null) {
```
Filter94: I think very soon it won't fit the case when there's a Maru without Besu EL instance

jvm-libs/test-utils/src/main/kotlin/maru/test/extensions/MaruAppExtensions.kt
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright Consensys Software Inc.
+ *
+ * This file is dual-licensed under either the MIT license or Apache License 2.0.
+ * See the LICENSE-MIT and LICENSE-APACHE files in the repository root for details.
+ *
+ * SPDX-License-Identifier: MIT OR Apache-2.0
+ */
+package maru.test.extensions
```
Filter94: We already have `testutils.maru.MaruExtensions`

jvm-libs/test-utils/src/main/kotlin/maru/test/genesis/BesuGenesisFactory.kt
```diff
@@ -0,0 +1,185 @@
...
+      return updatedGenesis
+    }
+
+    fun createGenesisWithClique(
+      genesisTemplate: String = genesisTemplateLondonWithoutConsensus,
+      cliqueBlockTimeSeconds: UInt,
+      cliqueEmptyBlocks: Boolean = true,
+      forks: ForksSchedule,
+    ): String {
+      var terminalTotalDifficulty: ULong? =
```
Filter94: Can be a val

jvm-libs/test-utils/src/main/kotlin/maru/test/genesis/BesuGenesisFactory.kt
```diff
@@ -0,0 +1,185 @@
...
+      cachedGenesis!!
+    }
+  }
+
+  companion object {
+    val genesisTemplateLondonWithoutConsensus: String =
+      GenesisConfigurationFactory.readGenesisFile("/besu-genesis-template.json")
+    private val jsonObjectMapper: ObjectMapper = jacksonObjectMapper()
+
+    fun createGenesisWithClique(
```
Filter94: I suggest to redesign this slightly:
1. make this method private
2. Make this method to return a JsonNode instead of a String

Then we can treat `JsonNode` as a Builder object and make it look even more like a builder if we change turn `setGenesisConfigProperty` into an extension method and make it return a JsonNode object

jvm-libs/test-utils/src/main/kotlin/maru/test/genesis/BesuGenesisFactory.kt
```diff
@@ -0,0 +1,185 @@
...
+      cliqueNode.put("blockperiodseconds", blockTimeSeconds.toLong())
+      cliqueNode.put("epochlength", blockTimeSeconds.toLong())
+      cliqueNode.put("createemptyblocks", createEmptyBlocks)
+
+      configNode.set<ObjectNode>("clique", cliqueNode)
+
+      return jsonObjectMapper.writeValueAsString(rootNode)
+    }
+
+    internal fun setGenesisConfigProperty(
```
Filter94: Why is this internal?

jvm-libs/test-utils/src/main/kotlin/maru/test/genesis/GenesisFactory.kt
```diff
@@ -0,0 +1,55 @@
...
+          validators = sequencersAddresses,
+          terminalTotalDifficulty = terminalTotalDifficulty,
+          forks = chainForks,
+        )
+    besuGenesisFactory.setForkSchedule(forkSchedule)
+    println("Initialized fork schedule: $forkSchedule")
+  }
+
+  fun besuGenesis(): String {
+    if (!this::forkSchedule.isInitialized) {
```
Filter94: I don't really see why BesuGenesis can't be accessed before Maru one

jvm-libs/test-utils/src/main/kotlin/maru/test/cluster/MaruCluster.kt
```diff
@@ -0,0 +1,313 @@
...
+      .filter {
+        it.nodeRole.isBootnode()
+      }.map { it.maru.p2pNetwork.enr!! }
+
+  fun start(): MaruCluster {
+    if (runningState != RunningState.STOPPED) {
+      throw IllegalStateException("Cannot START MaruCluster that is $runningState")
+    }
+    this.runningState = RunningState.STARTING
+    val nodesStartConfigs = nodesBuilders.map { it.build(genesisFactory::besuGenesis, besuCluster) }
```
Filter94: I think for Besu Genesis we should be able to store it as a String and access it directly in this class instead of the factory call

jvm-libs/test-utils/src/main/kotlin/maru/test/genesis/BesuGenesisFactory.kt
```diff
@@ -0,0 +1,185 @@
...
+          ?.let {
+            (it.configuration as DifficultyAwareQbftConfig).terminalTotalDifficulty
+          } ?: 0UL
+
+      var shanghaiTimestamp: ULong? = null
+      var cancunTimestamp: ULong? = null
+      var pragueTimestamp: ULong? = null
+      forks.forks.forEach { forkSpec ->
+        when (forkSpec.configuration.fork.elFork) {
+          ElFork.Paris -> {} // nothing to do, terminalTotalDifficulty already set
```
Filter94: I think it's not necessarily set. But if it's not set and the first fork is Paris, it should work fine

jvm-libs/test-utils/src/main/kotlin/maru/test/genesis/MaruGenesisFactory.kt
```diff
@@ -0,0 +1,111 @@
...
+          configuration = initialForkConfig,
+        )
+      return ForksSchedule(chainId, listOf(initialForkSpec))
+    }
+
+    val forkSpecks =
+      forks.entries
+        .sortedBy { it.key }
+        .also {
+          require(it.firstOrNull()?.key?.epochSeconds == 0L) {
```
Filter94: Minor: That's unnecessary strict. I believe first timestamp has to be lower than the current time, but it doesn't have to be 0

jvm-libs/test-utils/src/main/kotlin/maru/test/util/NetworkUtil.kt
```diff
@@ -6,18 +6,29 @@
  *
  * SPDX-License-Identifier: MIT OR Apache-2.0
  */
-package maru.p2p.testutils
+package maru.test.util
```
Filter94: This move means that P2P tests have to pull a lot of stuff from Besu, for example `org.hyperledger.besu.internal:besu-acceptance-tests-dsl`, which they don't need. I think `findFreePorts` belongs to P2P test fixtures

Filter94 (commented):
I didn't review all the tests yet

jvm-libs/test-utils/src/main/kotlin/maru/test/cluster/MaruCluster.kt
```diff
@@ -0,0 +1,313 @@
...
+    ),
+  val maruConfigTemplate: MaruConfig = configTemplate,
+  val maruClusterDataDir: Path = createClusterDir(),
+  val besuClusterWaitPeerDiscovery: Boolean = false,
+  val besuCluster: BesuCluster = BesuCluster(),
+) {
+  val genesisFactory: GenesisFactory = GenesisFactory(chainId, blockTimeSeconds)
+  private val nodesBuilders: MutableList<NodeBuilder> = mutableListOf()
+  val nodes = mutableListOf<ClusterNode>()
+  var runningState: RunningState = RunningState.STOPPED
```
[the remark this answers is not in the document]
fluentcrafter: Yes. For thread safety then the whole class needs to me, not just start/stop. 
I thought about make it threadsafe but it was not the priority. Anyway can easily achieve this by adding @synchronized  to the plublic methed. Will do that.

jvm-libs/test-utils/src/main/kotlin/maru/test/cluster/MaruCluster.kt
fluentcrafter: 1. it's meant to be private.
2. It's still WIP because I want to optimise the cluster start, but I'm having some chicken/egg issues.

jvm-libs/test-utils/src/main/kotlin/maru/test/cluster/NodeBuilder.kt
fluentcrafter: Yes, I am aware of that. Hence, I wanted the support for Maru without EL. Anyway, it should not be a blocker IMO, and we can address it in a follow-up PR.

jvm-libs/test-utils/src/main/kotlin/maru/test/extensions/BesuNodeExtensions.kt
```diff
@@ -0,0 +1,34 @@
...
+
+import linea.kotlin.toULong
+import maru.test.cluster.BesuCluster
+import org.hyperledger.besu.tests.acceptance.dsl.node.BesuNode
+import org.web3j.protocol.core.DefaultBlockParameter
+import org.web3j.protocol.core.methods.response.EthBlock
+
+fun BesuCluster.nodeHeads(): Map<String, ULong> = this.nodes.mapValues { it.value.latestBlockNumber() }
+
+fun BesuNode.latestBlock(includeTransactions: Boolean = false): EthBlock.Block =
```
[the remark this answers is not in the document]
fluentcrafter: > Or your intention is to change the testutils module's package layout?

The intention was to deprecate the other package and start using this one more organized...

jvm-libs/test-utils/src/main/kotlin/maru/test/extensions/MaruAppExtensions.kt
fluentcrafter: Same comment as above. I rather address those inconsistencies on follow up to avoid bigger diif and merge conflicts

jvm-libs/test-utils/src/main/kotlin/maru/test/cluster/MaruCluster.kt
fluentcrafter: The plan is to allow updating the list of forks in the cluster which updates the factory, then newly added nodes will get the updated genesis. So I rather keep it as is for now.

jvm-libs/test-utils/src/main/kotlin/maru/test/util/NetworkUtil.kt
fluentcrafter: > I think this way we're testing a subset of functionality (P2P) with a whole application, which is a more expensive way to test P2P than using P2P directly.

Define "expensive"

I care about: 1) readability, 2) maintenance, and 3) test runtime

Using this cluster with fake EL IMO will favour 1 and 2, with a negligible hint on 3.


Are you keen on this revert? The reason I am asking is that I foresee the `findPorts` being useful on something else that does not need the P2P module, and  honestly, I think we have other things with much higher priority and tech debt than squeezing perfect modules when we don't reach unanimity;

jvm-libs/test-utils/src/main/kotlin/maru/test/cluster/TestMaruAppFactory.kt
```diff
@@ -0,0 +1,37 @@
...
+  bootnodes: List<String> = emptyList(),
+  staticpeers: List<String> = emptyList(),
+  nodeKeyData: PrivateKeyGenerator.KeyData,
+  nodeRole: NodeRole,
+  forkSchedule: ForksSchedule,
+): MaruApp {
+  initPersistence(config.persistence, nodeKeyData)
+  var effectiveConfig = config
+  effectiveConfig =
+    setQbftConfigIfSequencer(effectiveConfig, isSequencer = nodeRole == NodeRole.Sequencer, nodeKeyData)
```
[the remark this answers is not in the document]
fluentcrafter: I prefer to keep the way it is because usually extension setX mutates the state and returns the same instance in a builder instead of returning a new instance, which is not the case with the data classes..

jvm-libs/test-utils/src/main/kotlin/maru/test/util/NetworkUtil.kt
```diff
@@ -6,18 +6,29 @@
...
+  fun findFreePorts(count: Int): List<UInt> {
+    val ports = mutableListOf<UInt>()
+    while (ports.size < count) {
+      val freePort = findFreePort()
+      if (!ports.contains(freePort)) {
+        ports.add(freePort)
+      }
+    }
+    return ports
+  }
```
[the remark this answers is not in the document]
fluentcrafter: A new implementation was introduced to mitigate this issue.

jvm-libs/test-utils/src/main/kotlin/maru/test/cluster/MaruCluster.kt
```diff
@@ -0,0 +1,326 @@
...
+    if (elNode is BesuElNode) {
+      besuCluster.addNodeAndStart(elNode.besu, awaitPeerDiscovery = besuClusterWaitPeerDiscovery)
+    }
+  }
+
+  private fun assertUniqueLabel(nodeLabel: String) {
+    val allNodesLabels = nodesBuilders.map { it.nodeLabel } + nodes.map { it.label }
+    require(allNodesLabels == allNodesLabels.distinct()) {
+      "Node label '$nodeLabel' is already used in the cluster. Please use unique labels for each node."
+    }
```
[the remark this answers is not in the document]
fluentcrafter: Good catch, it was a refactor regression. Pushed a fix and added a test. Thanks

jonesho (dismissed):

jvm-libs/test-utils/src/main/kotlin/maru/test/genesis/MaruGenesisFactory.kt
```diff
@@ -0,0 +1,114 @@
...
+    chainId: UInt,
+    validators: List<ByteArray>,
+    blockTimeSeconds: UInt = 1u,
+    forks: Map<Instant, ChainFork> = emptyMap(),
+    terminalTotalDifficulty: ULong? = null,
+  ): ForksSchedule {
+    require(blockTimeSeconds in 1u..60u) { "blockTimeSeconds must be between 1 and 60 seconds" }
+    require(validators.isNotEmpty()) { "Validators must not be empty" }
+    require(terminalTotalDifficulty == null || terminalTotalDifficulty > 0uL) {
+      "terminalTotalDifficulty must be greater than 0"
```
gauravahuja: terminalTotalDifficulty must be greater than 0 or null

gauravahuja (commented):
lgtm, just a few comments

jvm-libs/test-utils/src/main/kotlin/maru/test/cluster/TestMaruAppFactory.kt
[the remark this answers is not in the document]
Filter94: Yes, that's the point. We don't want to keep the old instance anyway, but it will improve the encapsulation. A `var` and `internal` are the symptoms of a weak design and my suggestion prevents them

jvm-libs/test-utils/src/main/kotlin/maru/test/genesis/BesuGenesisFactory.kt
Filter94: Nevermind, that was irrelevant

jvm-libs/test-utils/src/main/kotlin/maru/test/cluster/MaruCluster.kt
```diff
@@ -0,0 +1,313 @@
...
+  STOPPING,
+  STOPPED,
+}
+
+fun createClusterDir(): Path =
+  Files.createTempDirectory("maru-cluster-${Random.nextBytes(4).encodeHex(false)}").also {
+    it.toFile().deleteOnExit()
+  }
+
+class MaruCluster(
```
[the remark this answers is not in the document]
fluentcrafter: This is intended to mitigate the need for **repeated boilerplate code**, allowing the team to write tests faster and more consistently. Not an extremely opinionated framework with perfect care for encapsulation. 

That being said, here are some potential use cases:
 - besuCluster - in case someone needs to do something with a Besu node, e.g stop the node, clean the DB and restart from scratch with the same maru instance.
 - blockTimeSeconds,  maruConfigTemplate, maruClusterDataDir, chainId, ... - in case a test needs to read their values for whatever reason. This avoids creating extra vars in the tests
 - genesisFactory is private

Also, a test may need to create 1 or more clusters and read the config values of the 1st one when creating the 2nd one is handy to avoid create exta vars to hold some of those fiedls.

jvm-libs/test-utils/src/main/kotlin/maru/test/cluster/TestMaruAppFactory.kt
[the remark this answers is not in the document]
fluentcrafter: I am not following the comment...

Filter94 (dismissed):

jvm-libs/test-utils/src/test/kotlin/maru/test/MaruClusterTest.kt
```diff
@@ -0,0 +1,244 @@
...
+
+  @AfterEach
+  fun afterEach() {
+    if (::cluster.isInitialized) {
+      cluster.stop()
+    }
+  }
+
+  @Test
+  @Order(1)
```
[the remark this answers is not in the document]
fluentcrafter: It's because lack of resources and flakiness on the CI. In parallel, it would be too many Besu's at the same time and they suffer from starvation...

Filter94 (approved):

Issue #456, opened 2025-11-03
MaruAppCli testability


Filter94:
## Description
In [this](https://github.com/Consensys/maru/pull/453) PR there was some effort to test CLI parsing and some basic validation of the input. We should make MaruAppCli more testable

## Motivation
We currently don't have tests:
* for the multiple `--config` that override each other
* that ensure the order of the overrides, relative to `--config` files order

## Tasks
- [ ] Make MaruAppCli testable. For example we can pass an injectable `MaruAppFactory` which can be used in the test to verify the parsed configurations

## Acceptance criteria
- [ ] MaruAppCli is covered with the tests that check the overridability of the `--config` files
- [ ] MaruAppCli is covered with tests that check if the parsing of the genesis files specified in a different way

## Risks
- [ ] IDK

## Remember to
 - [ ] Add the `documentation` label in case there is an impact on the documentation
 - [ ] Add `priority` and `team` labels
 - [ ] Add Task for updating the Runbook or adding/updating existing metrics and alerts.

PR #457, opened 2025-11-04
feat: add test for multiple overrides config files

jonesho:
This PR implements #456

---

> [!NOTE]
> Add multi-file config support to the CLI, introduce a closeable lifecycle interface and factory abstraction, and update tests accordingly.
> 
> - **CLI**:
>   - Enhance `--config` to accept multiple files via comma-separated values or repeated options; update help text and parsing (`split=","`, `arity="1..*"`).
>   - Default network handling preserved; add `mixinStandardHelpOptions`.
>   - Use injected `MaruAppFactoryCreator` instead of constructing `MaruAppFactory` directly.
>   - On shutdown, call `app.stop()` then `app.close()`; remove `dryRun` and related checks.
> - **Lifecycle/Interfaces**:
>   - Introduce `LongRunningCloseable` (`LongRunningService` + `AutoCloseable`).
>   - Make `MaruApp` implement `LongRunningCloseable`; mark `start/stop` as overrides.
> - **Factory**:
>   - Extract `MaruAppFactoryCreator` interface; `MaruAppFactory.create(...)` now satisfies it and returns `LongRunningCloseable`.
> - **Tests**:
>   - Overhaul `MaruAppCliTest` to use temp files and a fake `MaruAppFactoryCreator`.
>   - Add coverage for multiple config files (both comma-separated and repeated) with override semantics; adjust existing tests to new paths and behaviors.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f7bc919a7ab74fede92be1a3d579319ec9f01f31. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/src/main/kotlin/maru/app/MaruAppFactory.kt
+++ b/app/src/main/kotlin/maru/app/MaruAppFactory.kt
@@ -133,7 +133,7 @@ class MaruAppFactory : MaruAppFactoryCreator {
       P2PState,
       () -> SyncStatusProvider,
     ) -> P2PNetworkImpl,
-  ): LongRunningCloseable {
+  ): MaruApp {
     log.info("configs={}", config)
     log.info("beaconGenesisConfig={}", beaconGenesisConfig)
 
--- a/jvm-libs/test-utils/src/main/kotlin/maru/test/cluster/TestMaruAppFactory.kt
+++ b/jvm-libs/test-utils/src/main/kotlin/maru/test/cluster/TestMaruAppFactory.kt
@@ -33,5 +33,5 @@ fun createMaru(
   return MaruAppFactory().create(
     config = effectiveConfig,
     beaconGenesisConfig = forkSchedule,
-  ) as MaruApp
+  )
 }
--- a/jvm-libs/test-utils/src/main/kotlin/testutils/maru/MaruFactory.kt
+++ b/jvm-libs/test-utils/src/main/kotlin/testutils/maru/MaruFactory.kt
@@ -377,7 +377,7 @@ class MaruFactory(
           }
         },
       p2pNetworkFactory = p2pNetworkFactory,
-    ) as MaruApp
+    )
 
   private fun buildP2pConfig(
     p2pPort: UInt = 0u,


app/src/test/kotlin/maru/app/MaruAppCliTest.kt
```diff
@@ -8,16 +8,234 @@
...
+      [syncing.download]
+      block-range-request-timeout = "10 seconds"
+      blocks-batch-size = 10
+      blocks-parallelism = 10
+      max-retries = 5
+      backoff-delay = "1 seconds"
+      use-unconditional-random-download-peer = false
+      """.trimIndent()
+
+    private val expectedMaruGenesisJson =
```
Filter94: I feel like this should be a minimal possible parseable genesis file, because its content is not essential for what we're testing. The parsing logic is thoroughly covered in our config tests

app/src/test/kotlin/maru/app/MaruAppCliTest.kt
```diff
@@ -8,16 +8,152 @@
...
+      private-key-path="./OVERRIDE/maru-db/private-key"
+
+      [payload-validator]
+      engine-api-endpoint = { endpoint = "http://OVEERRIDE:8550" }
+
+      [syncing]
+      peer-chain-height-polling-interval = "10 seconds"
+      """.trimIndent()
+
+    private val expectedMaruConfigDtoToml =
```
Filter94: Nit: It's not necessary to use a string here. You could have just instantiated the whole domain object instead. That way you make the assertion less dependent on the parsing logic, while keeping the copy-paste at the same level or lower, due to defaults set in the domain classes

app/src/main/kotlin/maru/app/MaruAppCli.kt
```diff
@@ -56,17 +56,22 @@ enum class Network(
   descriptionHeading = "%nDescription:%n%n",
   optionListHeading = "%nOptions:%n",
   footerHeading = "%n",
+  mixinStandardHelpOptions = true,
 )
 class MaruAppCli(
-  private val dryRun: Boolean = false,
+  private val maruAppFactory: MaruAppFactory = MaruAppFactory(),
```
Filter94: That's not a clean dependency injection without a separate interface for `MaruAppFactory`

The symptom of it is that`MaruAppFactory::create` returns an `LongRunningCloseable`, instead of a more specific type of `MaruApp`. This leads to `as MaruApp` type casts in `TestMaruAppFactory.kt` and `testutils.maru.MaruFactory::buildApp`, for example

app/src/test/kotlin/maru/app/MaruAppCliTest.kt
```diff
@@ -8,16 +8,234 @@
...
+    fun afterAll() {
+      listOf(tempMaruConfigFile, tempMaruConfigOverridesFile, tempMaruGenesisFile).forEach {
+        if (it.exists()) {
+          it.delete()
+        }
+      }
+    }
+  }
+
+  private val mockMaruAppFactory = Mockito.mock(MaruAppFactoryCreator::class.java)
```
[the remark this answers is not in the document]
Filter94: > spend more efforts to use fake/test-doubles?

Yes, the main motivation is to move away from the mocks in favor of test doubles.

> Can u be more specific on how to use a lambda here?

Currently in main `MaruAppFactory` has only 1 useful method, `create`. This is a sign that this signature can be easily turned into a functional interface and this enables us to just pass it on as lambda `{ NoOpMaruAppFactory }` or a test double that captures the parsed configs. If we choose to capture the config objects, you will probably need to pass a test double instead of a lambda. If you just want to return something specific, without assertions on the call arguments, lambda usually is a way to go

